Issue #8, opened 2020-12-31
Provide more deployment options


mgogoulos:
Deployment through the installation script is possible since the project has been released. 
However a list of other options should be facilitated. 

- [ ] Docker Compose
- [ ] Deploying to Heroku
- [ ] Deploying to Digital Ocean/AWS/Azure/GCP

gharsallahmoez:
@mgogoulos could you please tell us when the docker-compose deployment will be ready ?

mgogoulos:
Hello @gharsallahmoez , docker-compose deployment should be ready until February 2021, unless of course there's someone form the community helping out to bring this earlier!

gharsallahmoez:
@mgogoulos  could you please provide manual steps install of mediacms? because install.sh runs multiple systemd services and Docker is designed around the idea of a single service/process per container.

mgogoulos:
Sure, over the next few days I can prepare a wiki page documenting a few things that will facilitate the Dockerization of MediaCMS

swiftugandan:
I am in the process of creating a docker-compose file based on the design below: This design should act as a basis for a scaleable deployment in docker, kubernetes or docker swarm.

For this to work correctly, we shall need to remove any SSL setup within the 'built-in' nginx server, and delegate it to an ingress reverse proxy.

What are your thoughts around dropping SSL setup, and instead delegating it to a reverse proxy?

                                                                                                  
                                                                                                  
                                                                                                  
        +---------------------------------------------------------------+                         
        |                                                               |                         
        |                             Clients                           |                         
        |                                                               |                         
        |                                                               |                         
        +--------------------------------^------------------------------+                         
                                         |                                                        
                            +------------|-----------+                                            
                            |  nginx reverse proxy   |                                            
                            |  (ssl termination,     |                                            
                            |  letsencrypt client)   |                                            
                            +------------|-----------+                                            
                                         |                                                        
        +--------------------------------v------------------------------+    +---------------+    
        |                                                               |    |               |    
        |                     mediacms_web                              |    |               |    
        |                     (1 or many instances)                     ------               |    
        |                                                               |    |               |    
        +-----|--------------------------------------------|------------+    |               |    
              |   +--------------+                         |                 |               |    
              |   |  celery_beat |           +-------------|------------+    |               |    
              |   |  (singleton  -------------                          |    |               |    
              |   |  instance)   |           |       redis              |    |File store     |    
              |   +--------------+           |       (singleton         |    |(static_store, |    
[...]

PR #30, opened 2021-01-20
Deploy using docker compose

swiftugandan:
* Creates a docker-compose based deployment #8, see: [README](https://github.com/swiftugandan/mediacms/tree/deploy-using-docker-compose/deploy/docker)


mgogoulos:
Well done! This will need testing and a few issues to be solved before it can be merged, but looks as neat and high quality work.

* How can this be tested? Is docker-compose up, run as root, enough? Can you add a few notes, on deploy/docker/README.md, and perhaps move this to /docs and rename it as something like Docker_installation.md ?
Should this be run as user or root? What version of Docker compose do you use? Initially I tried it with 1.17 (Ubuntu 18.04 default) and got 

```
docker-compose up
ERROR: The Compose file './docker-compose.yaml' is invalid because:
services.web.build contains unsupported option: 'target'
```

With a newer version of Docker compose it proceeds to the point it hits this error:

```
Pulling celery_beat (mediacms:latest)...
ERROR: The image for the service you're trying to recreate has been removed. If you continue, volume data could be lost. Consider backing up your data before continuing.

Continue with the new image? [yN]ERROR: pull access denied for mediacms, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

(i hit y and it fails too). So it should be documented what is the minimum version of Docker compose this need to be run...

* I'm cautious about removing https for the systemd/local_install installation, as there are systems that run MediaCMS in place and this will create problems to them, if settings.SSL_FRONTEND_HOST stops being in use. While this mechanism of specifying  http/https connection is not ideal, and for sure it could be refactored in the future, it gives the chance to a default systemd/local_install  installation to function both with http + https out of the box, so if we can avoid removing this it would be great. Eg could we just let something as 
```
if local_install:
    SSL_FRONTEND_HOST = FRONTEND_HOST.replace("http", "https")
else:
    SSL_FRONTEND_HOST = FRONTEND_HOST

``` 
and add a note to refactor this code in the future (in all references where SSL_FRONTEND_HOST is used). 

So avoid changing the use of SSL_FRONTEND_HOST inside python code (context_processors, feeds, views etc), let nginx files as they are, just move from deploy/ to deploy/local_install, and allow both ways to run the software?

swiftugandan:
Will give a detailed response later, but for now you need to run
(docker-compose build) first before running (docker-compose up). Didn't
want to publish an image to the docker repo yet

swiftugandan:
Tested local_install and it is broken, systemd is not properly setup for envvar based overrides.

I will look at reverting local_install changes.

mgogoulos:
I run `docker-compose build && docker-compose up ` as root and all containers started, however since FRONTEND_HOST on docker-compose.yaml is http://mediacms.io all links would direct there. 

Once I changed it to http://localhost, everything run ok. Tried uploading a few files and they got uploaded/encoded well!

I don't think there's a need for an official docker image now, it's great that the code is mounted so it can also be easily updated.

 I have uploaded a few medias on my local installation and notice that media_files directory is not mounted on the container. I think this is important if the same directory is mounted, because it will offer an easy way to a) get all media files, b) export database, and this would be very helpful on home/small installations that are greatly facilitated by the containers, so users should have an easy way to get a full backup.  The same applies for other things, even logs, easy access if logs are on logs/ rather than on each container itself...

Please look if you can revert changes to local installation so this is not affected!

swiftugandan:
@mgogoulos pushed some updates hopefully its all working. Please test and let me know ...

mgogoulos:
- [ ] Local installation (localhost + remote server) seem unaffected
- [ ] Docker installation at localhost: installs and runs ok, however I had a problem uploading content due to permissions on media_files. The container didn't have write access there. I think what needs to be done here is the same as in postgres_data
- [ ] Docker installation on remote server: install, but links point to localhost. Were should I change the FRONTEND_HOST variable? Editing settings.py file has no effect I see. I need to see how easily I can a) edit these settings and see effect, b) if I edit content in static/ , how to see changes take effect. For example, one of the first user actions is usually to set a new logo. 
- [ ] If I understand well, when docker-compose build runs, it copies the source inside the container? How about instead mounting the source code, so that any changes made on the code take effect (without the need to run docker-compose build again? The idea here is that if I clone the repository, run docker-compose inside it, and have the system running, how do I see changes to the code?
- [ ] same with logs/ , instead of having each container writing logs to their logs/ directory, if the whole directory is mounted, then would all logs be written in logs/ dir? Then it would be easy to handle with logrotate, debug etc...

swiftugandan:
I have pushed an update that hopefully fixes most of the issues mentioned above.

* The default docker-compose.yaml will use the current source for runtime. After running `docker-compose build` at least once.
* You can change the FRONTEND_HOST in deploy/docker/local_settings.py then restart the affected containers for the change to take effect.
* Persisted data like media_files, static resources, logs, etc have been git ignored.
* The postgres_data has been placed a level outside the source

Please take note it is not recommended to use mounts for production. Named volumes are preferred. See: https://docs.docker.com/storage/volumes/ and https://docs.docker.com/storage/bind-mounts/

On my dev machine, I have experienced reduced performance while using mounts.

gharsallahmoez:
@swiftugandan it doesn't work on my server. I try it with docker-compose.yaml and docker-compose-https-proxy.yaml, I added my certificate in deploy/docker/reverse_proxy/certs and I added my server in ALLOWED_HOSTS in cms/settings.py . I get `(failed)net::ERR_CONNECTION_REFUSED` when I inspect the network in my browser. 
is there any other required configuration I need to add it?

swiftugandan:
> @swiftugandan it doesn't work on my server. I try it with docker-compose.yaml and docker-compose-https-proxy.yaml, I added my certificate in deploy/docker/reverse_proxy/certs and I added my server in ALLOWED_HOSTS in cms/settings.py . I get `(failed)net::ERR_CONNECTION_REFUSED` when I inspect the network in my browser.
> is there any other configuration I need to add ?

1. In `deploy/docker/local_settings.py` change `FRONTEND_HOST` to your host, eg. `https://www.example.com`
2. Add certs named `www.example.com.crt` and `www.example.com.key` to `deploy/docker/reverse_proxy/certs/`
3. In `docker-compose-https-proxy.yaml` update `VIRTUAL_HOST: www.example.com`
4. Then run `docker-compose -f docker-compose-https-proxy.yaml up`

Hopefully that works for you.

gharsallahmoez:
@swiftugandan no it doesn't, maybe because I'm not setting www CNAME in my server, should I add it ?

swiftugandan:
> @swiftugandan no it doesn't, maybe because I'm not setting www CNAME in my server, should I add it ?

This assumes you have already configured your DNS to point www.example.com to your servers IP address, and port 80 and 443 are open on the server.

mgogoulos:
Hi @swiftugandan , this is great work and I don't want to have it blocked, however at this point I cannot merge it. Even if all different docker setups are not tested extensively now, I'm willing to merge it, as long as a) the local installation works as it is now, b) the docker installation works for localhost. 

I have tried the installation on a remote server (install.sh), it completes, but all links point to localhost. Seems that cms/local_settings.py was copied to my cms/local_settings and the settings there took effect. Also I saw this

```
DEBUG = FalseFRONTEND_HOST='test.mediacms.io'
```
so some redirect didn't work well. But most important, cms/local_settings.py should be on .gitignore, and not on the version control...I see you're using it for the docker installation so it shouldn't interfere with the local installation. 

Then there are two small fixes and one suggestion: a) the SECRET_KEY should not be hard coded, and b) the same for the ADMIN_PASSWORD. Both could introduce unnecessary issues. Can you make sure they are dynamic? 

Then, on README.md, below Install and Update sections, can you enter a `Docker install` section, where you document the same things you do on another place (just the minimum versions of Docker + Docker-compose, the two commands that are needed to be run as root/sudo, plus a link for more info to the doc you've created)

Thank you very much!

gtntdev:
Hi @mgogoulos , 
I tested every commit (even the latest) of the docker version with a slightly modified Traefik configuration.
Used over 30GB of video data and never encountered any errors.
If you want I can send you the url via PM (server is live).

In the latest configuration i misconfigured the host (I copied a trailing '/' but it's working anyway, just hat no time to rebuild).
A few hints on the functionality:
- the "playlist create" button isn't very obvious, maybe it should be displayed in the playlist sectiont (user's "My playlist") from the menue too.
- you can't add private playlists (private videos to playlists) and you can't share videos between users (private)
These features would be nice to have but thanks for the great work you did so far!

If you want access to my server or if I can help you in any other way just let me know.
I'd really like to see this PR merged:)

Best regards,
Lukas

swiftugandan:
> Hi @swiftugandan , this is great work and I don't want to have it blocked, however at this point I cannot merge it. Even if all different docker setups are not tested extensively now, I'm willing to merge it, as long as a) the local installation works as it is now, b) the docker installation works for localhost.
> 
> I have tried the installation on a remote server (install.sh), it completes, but all links point to localhost. Seems that cms/local_settings.py was copied to my cms/local_settings and the settings there took effect. Also I saw this
> 
> ```
> DEBUG = FalseFRONTEND_HOST='test.mediacms.io'
> ```
> 
> so some redirect didn't work well. But most important, cms/local_settings.py should be on .gitignore, and not on the version control...I see you're using it for the docker installation so it shouldn't interfere with the local installation.
> 
> Then there are two small fixes and one suggestion: a) the SECRET_KEY should not be hard coded, and b) the same for the ADMIN_PASSWORD. Both could introduce unnecessary issues. Can you make sure they are dynamic?
> 
> Then, on README.md, below Install and Update sections, can you enter a `Docker install` section, where you document the same things you do on another place (just the minimum versions of Docker + Docker-compose, the two commands that are needed to be run as root/sudo, plus a link for more info to the doc you've created)
> 
> Thank you very much!

* Did you run `install.sh` on the same remote machine you had run the docker installation?

swiftugandan:
@mgogoulos I would like to use https://pypi.org/project/python-decouple/ to capture settings from the user via a settings.ini file for .env file, instead of letting them edit the local_settings.py (for docker deployment, but could also be extended to the local_install)

What are your thoughts?

mgogoulos:
> Hi @mgogoulos ,
> I tested every commit (even the latest) of the docker version with a slightly modified Traefik configuration.
> Used over 30GB of video data and never encountered any errors.
> If you want I can send you the url via PM (server is live).
> 
> In the latest configuration i misconfigured the host (I copied a trailing '/' but it's working anyway, just hat no time to rebuild).
> A few hints on the functionality:
> 
> * the "playlist create" button isn't very obvious, maybe it should be displayed in the playlist sectiont (user's "My playlist") from the menue too.
> * you can't add private playlists (private videos to playlists) and you can't share videos between users (private)
>   These features would be nice to have but thanks for the great work you did so far!
> 
> If you want access to my server or if I can help you in any other way just let me know.
> I'd really like to see this PR merged:)
> 
> Best regards,
> Lukas

Hi Lukas, good to read that you've tried this out and encountered no issues. Can you open an issue with the findings related to the playlists and another one with the private video files? So we can handle them. 
Thanks!

mgogoulos:
> @mgogoulos I would like to use https://pypi.org/project/python-decouple/ to capture settings from the user via a settings.ini file for .env file, instead of letting them edit the local_settings.py (for docker deployment, but could also be extended to the local_install)
> 
> What are your thoughts?

Sounds good! Do you prefer to work on it on this PR or on another one?

mgogoulos:
This is good enough to go, as long as others have tested it too. Any improvements can take place in the near future - thanks very much!

swiftugandan:
> This is good enough to go, as long as others have tested it too. Any improvements can take place in the near future - thanks very much!

Will make improvements with time, quite busy with work at the moment

PR #55, opened 2021-02-16
fixes hls generation for docker builds

swiftugandan:
* Had trimmed out too much from the bento-sdk
* placed the bento a level below the source, so it is not mounted over by the source folder

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 media_files/encoded/
 media_files/original/
+media_files/hls/
 postgres_data/
 celerybeat-schedule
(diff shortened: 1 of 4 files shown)


mgogoulos:
Just tried this on a new installation and it doesn't work - HLS files are not created. 

I've noticed thought that if I manually create the hls folder, eg `root@docker:~# mkdir  mediacms/media_files/hls ; chown www-data. mediacms/media_files/hls/` then it runs as expected.

swiftugandan:
Fixed

PR #69, opened 2021-03-02
use mediacms image from docker hub

mgogoulos:


docker-compose-http-proxy.yaml
```diff
@@ -27,7 +27,7 @@ services:
     build:
```
swiftugandan, line 27: build is not needed anymore

Issue #38, opened 2021-01-24
[Feature] Improve visibility on video encodings and allow to re-encode


mgogoulos:
- Provide logs (that are already stored) regarding encodings
- Provide option to re-run encoding tasks (for site editors/managers and admins)

PR #99, opened 2021-03-13
use django-db as celery results backend

swiftugandan:
* use django-db as celery results backend
* use django-db as celery-beat scheduler

These changes will help give more visibility into the executed tasks and task schedules. We shall also be able to monitor the workers using flower

### DIFF
--- a/cms/settings.py
+++ b/cms/settings.py
@@ -288,6 +288,9 @@ INSTALLED_APPS = [
     "uploader.apps.UploaderConfig",
     "djcelery_email",
     "ckeditor",
+    'celery',
+    'django_celery_results',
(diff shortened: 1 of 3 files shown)


swiftugandan:
related to https://github.com/mediacms-io/mediacms/issues/38

mgogoulos:
+1 for installing flower, but I'm reluctant to use django-celery-results

Quoting from https://vedantsopinions.medium.com/how-we-scaled-celery-for-our-django-app-da2465a3a6be

```
Storing task results in a centralized location (like SQL) defeats the fundamental tenet of celery: a distributed task queue.
```

PR #128, opened 2021-03-31
Update install.sh

sthierolf:
added check for Debian 10 (buster)
packages provided by buster have same naming like in ubuntu 20


swiftugandan (changes requested):

mgogoulos:
Hey sorry for replying to this late. Although docker-compose setup is the way I would now recommend for a new installation, there are still cases where one might consider installing via the single server setup, so I'm ok to merge this!

Can you make a change, on README.md, edit `Automated script - to run on Ubuntu 18 or Ubuntu 20 flavors only!` and make it include Debian buster as well?

mgogoulos:
ok merged, I'll perform that change now. Thanks!

PR #132, opened 2021-04-01
Set default theme in settings

Dan1ell:


mgogoulos:
nice catch! Merging @styiannis

PR #165, opened 2021-04-30
 adds drf-yasg and automated generation of Swagger Schemas

mgogoulos:
adds drf-yasg
updates Django to 3.1.8


mgogoulos:
this needs the following:

- [ ] test that it doesn't break something (specially the Django upgrade)
- [ ] add model definition to Swagger
- [ ] add parameter definition to Swagger for views

PR #171, opened 2021-05-04
Fix typo in settings.py

phreaknerd:
Guessing this "f" should not be there?


phreaknerd:
Closed this one and split it up into two requests for your convenience.

PR #172, opened 2021-05-04
Fix typo in settings.py

phreaknerd:
There was a leading f in front of a value in settings.py. Guess that was a typo.


swiftugandan:
This is valid syntax ... see: https://realpython.com/python-f-strings/

phreaknerd:
Ah sorry. Thanks for the advice!

PR #173, opened 2021-05-04
Add default local settings file

phreaknerd:
Create a default_local_settings.py as a template for local_settings.py.
We should support all beginners with a good workflow to keep their credentials save.

### DIFF
--- a/cms/default_local_settings.py
+++ b/cms/default_local_settings.py
@@ -1,0 +1,27 @@
+# This is a template for your local settingsfile.
+# Copy this file to local_settings.py and you're able to override
+# every setting from settings.py for your local setup.
+
+DEBUG = False
+
+# email settings
+DEFAULT_FROM_EMAIL = "info@mediacms.io"
+EMAIL_HOST_PASSWORD = "xyz"
+EMAIL_HOST_USER = "info@mediacms.io"
+EMAIL_USE_TLS = True
+SERVER_EMAIL = DEFAULT_FROM_EMAIL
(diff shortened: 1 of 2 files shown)


Dan1ell:
But if you want to support beginners with a good workflow, shouldn't you just explain good workflow in the docs then?

phreaknerd:
I did. Or is there another location for the docs? 
I'd suggest to create an own chapter for the setup of mediacms in context of deployment/updates/git and would be glad to support with that.
The existence of a "default_local_settings"-file is a common practice in other projects afaik (symfony-framework projects or Drupal) and triggers at least some research and questioning when working with the settings.

mgogoulos:
Hi, this looks good for the single server installation, but for the docker installation deploy/docker/local_settings.py it has already content there so if it gets overridden by this it will miss what's there. 

A quick solution for this PR to get merged is to remove mention on the docker installation on docs/Configuration.md 

But also maybe it's a good chance to make both ways consistent? In this case it seems that the single server installation has a cms/settings.py file and you can create a cms/local_settings.py file to override some settings there, without a problem if you update the software since cms/local_settings.py is not part of the repository. 
But on the docker installation, if you make changes on deploy/docker/local_settings.py and then update the software, there might be conflicts to what you have specified there and what has been committed to that file as it is part of the repository...

mgogoulos:
I'll close this since docs/Configuration.md does not exist anymore. @phreaknerd please re-open on new docs/ files if you want!

PR #178, opened 2021-05-08
adding docker letsencrypt example

swiftugandan:
An example that shows how you can do letsencrypt for a docker-based mediacms deployment.

In this example, I have used named volumes, but feel free to use your host volumes as in `docker-compose.yaml`


mgogoulos:
good to go! I will add a few notes on the Docker compose config page, thanks!

PR #179, opened 2021-05-11
Fix or Playback Issues for iPad + GET-Parameters (autoplay,muted,time) for Embed

DieProgrammIDE:
Hello,
these are changes i made for my own usage, fixing the blocking bug that caused the Player to not work on iPad and only play Audio (#5).
Also i added "playsinline" for the Player to play in windowed mode on iPhone.

Because it is needed in my use-case, the Player now allowed GET-Parameters for Autoplay, Muting and Start-Time.

Although the Code may not be clean it does work and at least can be seen as a workaround for others with similar problems in the meantime.


mgogoulos:
Awesome, thank you very much for submitting this! It's impressive you managed to provide a solution without the frontend code. Myself and @styiannis will have a look and reply soon with feedback!

styiannis:
@DieProgrammIDE If I understand correctly, in your use-case you pass some GET parameters from URL address and use them to configure video player behavior. If this is the case then maybe it's a good idea to separate the code in order we can merge in the official branch only the changes related to the bug fix and the definition of "playsline" option

DieProgrammIDE:
I can remove that functionality. Autoplay doesn‘t work reliably anyway, sometimes, play gets fired before the video is there.

mgogoulos:
Hi all, what's the status on this? @DieProgrammIDE thinking of making any changes here? @styiannis what do you think?

styiannis (approved):
Although some improvements could be made to the code, there is no problem so we move on to merging it. In the near future, the branch "[frontend-dev-env](https://github.com/mediacms-io/mediacms/tree/frontent-dev-env)" with the code and development  environment of the frontend will be merged into the main branch, so any relevant improvements will be made through it.

PR #218, opened 2021-06-09
Segregation of Dev and Prod envs

shubhank-saxena:
This pull request will start the addition of testing modules for the entire codebase.

The various packages needed for this task are -
- Pytest-Django - This is a modified pytest package useful to test Django projects
- Factory-Boy - This is a factory data generator package that will help us to generate data for tests
- Faker - A package that generates fake data for testing

Structural changes to the repo
- A folder named `tests` would be created in the project's root folder, which will have sub-folders named as the respective Django app names.
- Conftest file in the root will help us to declare all data factories in one place.
- Pytest.ini in root is a generic setting file for pytest and its behaviour.

@mgogoulos please let me know if this proposed method requires any further changes. Once we finalized this process, we can then proceed to start writing tests for every respective django app


shubhank-saxena:
I think we are good to go with this PR @mgogoulos 
I have opened another PR for UI tests

Dockerfile
```diff
@@ -55,6 +55,11 @@ RUN apt-get update -y && apt-get -y upgrade && apt-get install --no-install-reco
     apt-get purge --auto-remove && \
     apt-get clean
 
+RUN apt-get update && apt install -qqy gnupg2 wget unzip libnss3-dev libgdk-pixbuf2.0-dev libgtk-3-dev libxss-dev && \
```
swiftugandan, line 58: This docker file is optimized for production, we should not be adding test artifacts to it (these will bloat it, yet they are not needed for the runtime).

Perhaps we should create another one that uses this as the base

docker-compose.yaml
```diff
@@ -18,6 +18,7 @@ services:
         condition: service_healthy
   web:
     image: mediacms/mediacms:latest
+    build: .
```
swiftugandan, line 21: We should not be changing this docker-compose, it's used for production. Create another which uses the image (with test artifacts) mentioned above

requirements.txt
```diff
@@ -31,6 +31,13 @@ django-ckeditor
 
 django-login-required-middleware==0.6.1
 
+pytest-cov
```
swiftugandan, line 34: Again these are not needed in the production image, we need to separate prod requirements and dev/test requirements

swiftugandan (changes requested):

requirements.txt
```diff
@@ -31,12 +31,6 @@ django-ckeditor
...
 django-login-required-middleware==0.6.1
 
-pytest-cov
-pytest-django
-pytest-factoryboy
-Faker
-selenium
-webdriver-manager
 
 # extra nice utilities!
```
swiftugandan, line 35: All these are dev dependencies and should be placed in the requirements-dev.txt

docker-compose.dev.yaml
```diff
@@ -0,0 +1,40 @@
+version: "3"
+
+services:
```
swiftugandan, line 3: This is incomplete, below to the list of services:

```
.
.
.
.
  web:
    build:
      context: .
      dockerfile: ./Dockerfile.dev
    image: mediacms/mediacms:latest
    ports:
      - "80:80"
    volumes:
      - ./:/home/mediacms.io/mediacms/
    depends_on:
      redis:
        condition: service_healthy
      db:
        condition: service_healthy
  db:
    image: postgres
    volumes:
      - ../postgres_data:/var/lib/postgresql/data/
    restart: always
    environment:
      POSTGRES_USER: mediacms
      POSTGRES_PASSWORD: mediacms
      POSTGRES_DB: mediacms
    healthcheck:
      test: ["CMD-SHELL", "pg_isready -U mediacms"]
      interval: 10s
      timeout: 5s
      retries: 5
  redis:
    image: "redis:alpine"
    restart: always
    healthcheck:
      test: ["CMD", "redis-cli","ping"]
      interval: 30s
      timeout: 10s
      retries: 3
```

requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
+pytest-cov
```
swiftugandan, line 1: This extends requirements.txt, so add `-r requirements.txt` to the top of the file

You need to create a `Dockerfile.dev` that adds these requirements to a dev docker container

swiftugandan (changes requested):

docker-compose.dev.yaml
shubhank-saxena: Actually we don't need to write all of them again. If we simply run
`docker-compose -f docker-compose.yaml -f docker-compose.dev.yaml up` it will run docker-compose first and then overrite non-existant settings from docker-compose.dev and run them

swiftugandan: Good suggestion, I would use `extends` (https://docs.docker.com/compose/extends/#extending-services) instead of depending on the user getting the command line right

Also, take note:

* some services have been removed (in dev mode they will all run in the web container)
* the `Dockerfile.dev` is used to create the image for the web container in dev mode

docker-compose.dev.yaml
```diff
@@ -0,0 +1,54 @@
...
+  #     - ${PWD}/frontend:/home/mediacms.io/mediacms/frontend/
+  #   working_dir: /home/mediacms.io/mediacms/frontend/
+  #   command: bash -c "npm install && npm run start"
+  #   env_file:
+  #     - ${PWD}/frontend/.env
+  #   ports:
+  #     - "8097:8097"
+  #   depends_on:
+  #     - web
+  web:
```
swiftugandan, line 16: I think you still need to copy the db and redis service into this file. Using the chained `docker-compose -f file1 -f file2` will merge settings from file1 to file2. So the web service will get Environment variables that have intentionally been removed from here (so defaults are used)

Dockerfile.dev
```diff
@@ -0,0 +1,67 @@
+FROM python:3.8-buster AS compile-image
```
swiftugandan, line 1: Feels like this should extend the `mediacms/mediacms:latest` and since it is a dev image, it does not need a 2 stage build.

Let me check and revert

shubhank-saxena: Makes sense!

docker-compose.dev.yaml
shubhank-saxena: I don't understand this. So when we run chained command, it inherits non changed values. So won't that bring over redis and db?

swiftugandan: It will also bring in ENVIRONMENT in web, which has been removed

swiftugandan (approved):

swiftugandan (approved):

.github/workflows/python.yml
```diff
@@ -1,37 +1,28 @@
...
-      run: python manage.py migrate
-    - name: Run tests
-      run: py.test
+    - name: Checkout
+      uses: actions/checkout@v1
+
+    - name: Build the Stack
+      run:  docker-compose -f docker-compose-dev.yaml build
+
+    - name: Run DB Migrations
```
swiftugandan, line 21: I think you need this step instead of the migrations step...

```
    - name: Start containers
      run: docker-compose -f docker-compose-dev.yml up -d
```

Issue #214, opened 2021-06-05
Playing 2,7k videos (2720 × 1530) in original resolution (or new encoding profile)


DecaTec:
**Describe the feature you'd like**
I have some video footage in 2,7k (2720 × 1530). This gets encoded in all lower resolutions configured (e.g. 720p, 1080p). But when playing the video, the max. resolution then is 1080p.
I would expect that besides the encoded versions, the video will be played in original resolution as max. resolution. This doesn't seem to be the case.

So the feature request would be that all videos can be played in original resolution besides the encoded resolutions.

**Describe alternatives you've considered**
As an alternative, there should be an corresponding encoding profile for 2,7k (1530p). At the moment, this cannot be added in the admin UI under encoding profiles.

DecaTec:
Added PR for new encoding profiles in 2,7k: #219

PR #219, opened 2021-06-09
Add 2,7k encoding profiles

DecaTec:
## Description
Added 2,7k (2704x1520 and 2720 × 1530) resolution to encoding profiles.
These are **not active by default**.

2,7k are common resolutions on drones and action cams. But there doesn't seem to be a standard in resolution for 2,7k, so sometimes it's 2704x1520 and sometimes 2720 × 1530.

Tested on a Docker deployment. Rebuild container and then `docker-compose up`.
Not sure about the bitrate values in `files/helpers.py`, so maybe some fine tuning needs to be done.

See #214


mgogoulos:
Thanks for the PR! How did you find these bitrates, it would be great if you can share a link. 
Can you create a new migrations file? We should add new migration files rather than edit existing ones, mostly for backward compatibility

Please make the changes and I can merge it after performing a few tests

DecaTec:
Well, I'm really not into Python/Django development, but I hope that the latest commit ist correct regarding migrations.
Please have al look.

Honestly, the bitrates are just guessed (something in between 1440p and 4k). Does these need to be "correct"? Maybe you can point me to the right direction to determine the correct bitrates then?

mgogoulos:
1. the migrations files are ok, but I couldn't install the fixtures (with python manage.py loaddata fixtures/encoding_profiles.json). Did this work for you? Gives me an error 

```
json.decoder.JSONDecodeError: Invalid control character at: line 1 column 4096 (char 4095)

```

2. not sure whether this will work on the player - it's a good way to learn though! did you get it working on your side without having to do anything on the frontend?

DecaTec:
Sorry, still learning...

Started from a fresh install with my fork of the project.
Called `install.sh`.

After that, the new profiles appear in the encoding profile section of the admin UI.

Calling the command you mentioned gives me:
```
/home/mediacms.io/mediacms# python3 manage.py loaddata fixtures/encoding_profiles.json
Installed 28 object(s) from 1 fixture(s)
```

The error you get seems to be an invaliud char in the json file, but I cannot reproduce.

What can I do to reproduce/fix the problem?

Second point: Yes, the new encoding profiles are shown in the admin UI. I've tested these by only acivating the new profiles.

DecaTec:
Any news in this one?

mgogoulos:
Hi @DecaTec , sorry for the delay, I'll try to get this tested and merged very soon. On the meanwhile, can you upload 1-2 files of those dimensions, on https://demo.mediacms.io/ so I can better check? 

Thanks!

DecaTec:
Thanks for the update.

I've uploaded two videos to the demo instance:
* 1530p: https://demo.mediacms.io/view?m=Mxn5RCIL8
* 1520p: https://demo.mediacms.io/view?m=1OpAXrqmq

multiflexi:
Hi. I think profile resolutions should be based on the viewing device, the screen, not the resolution of the source material. So this source should be encoded in 1440 profile tops.

DecaTec:
Shouldn't the videos be played in highest possible quality (i.e. highest available resolution) as long `resolution video <= resolution device`?

However, these new profiles may not be commonly used, so they are disabled by default.

multiflexi:
Usually no. The resolution is based on the device which is used for the content consumption. All resolutions are based on the resolution of displays. HD 720/1080, UHD2160/4320 and 1440/2880 are all resolutions of displays, not what cameras output. Another aspect is compatibility because not all devices can decode all resolutions, for example, older GPUs/smartphones are limited to 1080, nowadays the mainstream limit is usually 2160, only the newest/high-end devices can do more. If the profiles were based on cameras, there would be quite a lot of them as there is a multitude of cameras with non-standard output resolution. A more complex solution would be to let users define the vertical resolution of profiles. Then everybody would be able to do whatever they want.

DecaTec:
@mgogoulos any chance that this will get merged? Otherwise we could just close this. What do you think?

mgogoulos:
Sorry for letting this hang for such a long time. I will close it and potentially re-open it in future, thanks a bunch for submitting!

PR #220, opened 2021-06-15
Webserver/setup optimizations

DecaTec:
## Description

Some optimizations for the webserver configuration regarding security. 
Also, the `setup.sh` is optimized for non standard nginx configurations (e.g. when installed from the official nginx repositories on Ubuntu).

## Changes
- Webserver/security:
    - Security: Do not use TLSv1.0 and TLSv1.1 -> TLSv1.2 and TLSv1.3 by default.
    - Added `secure ssl_ciphers` in nginx vrtual host.
    - Moved all the certificate and SSL options to the mediacms.io virtual host (these are not included in the `nginx.conf` anymore)
    - Added `ssl_ecdh_curve` option.
    - Created default DH params (4096 bit).
- Setup optimizations:
    - The directories `/etc/nginx/sites-available` and `/etc/nginx/sites-enabled` are created during setup (in newer nginx configurations, these directories are missing because the vHosts are usually stored under `/etc/nginx/conf.d/*.conf`).
    - The default vHost is not overwritten by the setup, only removed from `/etc/nginx/sites-enabled`
    - The vHost for MediaCMS is now not the default vHost, but `mediacms.io`.
    - The MediaCMS vHost in `sites-enabled` is now a symlink to the vHost under `sites-available`.
    - Automated regeneration of DH params (4096 bit) during setup (only if real certificates are used and MediaCMS is not running on localhost with self signed certificates).

## Tests
 Tested on:
 - "Single server installation"
 - Docker installation

 ## Additional information

 A good starting point for testing webserver secutity is [Qualys SSL Labs](https://www.ssllabs.com/ssltest/).


mgogoulos:
thanks for the changes @DecaTec , tested them on local installation and they are good!

We could bring them to the nginx configuration for the Docker Compose setup at some point

DecaTec:
Thanks for merging!

These changes mainly focus in the webserver security (SSL settings, etc.).
When I got this right, you'll need some reverse proxy in front of the installation when using docker-compose. The security related features should then be part of the reverse proxy configuration.

Issue #224, opened 2021-06-16
pids folder should exist only in container scope


swiftugandan:
If the deployment mounts the entire source folder, the pids folder is visible to all containers. This means that when any of the containers are starting, they will clean up any existing pids see:
https://github.com/mediacms-io/mediacms/blob/de30fe68f12ea9e75880b501d217e1c51911213d/deploy/docker/entrypoint.sh#L14 ... Looks like this was a hack around a bigger issue...

We need to place the pid files within the scope of an individual container (eg in /var/run/) to prevent containers from interfering with PID files of other containers

swiftugandan:
@mgogoulos this is resolved

mgogoulos:
Thanks @swiftugandan !

PR #225, opened 2021-06-17
keep pids in container scope

swiftugandan:
## Description
Stores celery pids within a running container state. See: https://github.com/mediacms-io/mediacms/issues/224#issue-923117670

## Steps

*Pre-deploy*
N/A

*Post-deploy*
N/A


mgogoulos:
tested on a few Dockerized scenarios and works well

PR #228, opened 2021-06-18
shutdown migrations container when migrations have completed

swiftugandan:
## Description
* Migrations are run as a pre-start process, there is no need to waste resources by unnecessarily keeping the container running when the migrations have been completed.
* This MR solves this issue

## Steps
*Pre-deploy*
N/A
*Post-deploy*
N/A


mgogoulos:
tested and works as expected, merging

Issue #118, opened 2021-03-29
docker-compose setup and migrations


Dan1ell:
During the update process 'docker-compose up' always loads the fixtures. But then the categories fixture keeps resetting my categories list. There should be a straightforward way to update MediaCMS without running the fixtures again. (Currently, I am commenting out ‘manage.py loaddata’ in prestart.sh).

Initial data should really only load initially. When some data is already present in the intended table or column, initial data should not be loaded again. But the user should be able to override this behavior with a variable in local_settings, perhaps for use in the dev environment, to intentionally reset test data.

Regarding the variable, "LOAD_INITIAL_DATA" might not be the right name for this yes/no variable since you pretty much always want to load initial data when it’s actually the first time it’s going to load; so on installation but also later whenever a new migration runs for the first time to add a model or add a field. If that change came with some initial data, you would want it. So the answer would always be ‘yes’, I want to load initial data.

A better name might be RELOAD_INITIAL_DATA because it is clear that setting this variable to ‘no’ does not prevent the first loading of fixture data, it’s about a reload, it loads the fixture data even if data is already there.

And I think if RELOAD_INITIAL_DATA is set to ‘yes’, it should replace the data that is already there not just concatenate to it like it’s doing now. Because I would use it to reset the test data. So maybe RESET_INITIAL_DATA is also good.

Because the way I see this being used is so on dev we can automatically start with fresh test data after every update and we can add our own fixtures and list them in the prestart.sh so that they are also controlled by this variable.

Or we can list our fixtures in local_settings and prestart.sh will be modified to look for this list in the settings.

(This issue came out of discussion #115)

swiftugandan:
We need to run the migrations script only once ... Perhaps have a check on the db level, eg check that Users exist ( from django.contrib.auth.models import User; User.objects.exists())and only run it if this is False?

mgogoulos:
the migrations service does other stuff as well that need be rechecked and only run initially

mgogoulos:
Fixed on https://github.com/mediacms-io/mediacms/pull/229 !

PR #229, opened 2021-06-18
optimize migrations service to not run all actions every time

mgogoulos:
## Description
addresses https://github.com/mediacms-io/mediacms/issues/118 , doesn't run all steps from prestart.sh every time migrations service starts but rather when DB is empty, and system is uninitialized

also sets a random password for admin (used to be a fixed pw) so update on documentation is needed

### DIFF
--- a/deploy/docker/prestart.sh
+++ b/deploy/docker/prestart.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 RANDOM_ADMIN_PASS=`python -c "import secrets;chars = 'abcdefghijklmnopqrstuvwxyz0123456789';print(''.join(secrets.choice(chars) for i in range(10)))"`
-ADMIN_PASSWORD=${RANDOM_ADMIN_PASS:-$ADMIN_PASSWORD}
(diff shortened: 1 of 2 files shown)


deploy/docker/prestart.sh
```diff
@@ -1,22 +1,31 @@
 #!/bin/bash
 
 RANDOM_ADMIN_PASS=`python -c "import secrets;chars = 'abcdefghijklmnopqrstuvwxyz0123456789';print(''.join(secrets.choice(chars) for i in range(10)))"`
-ADMIN_PASSWORD=${ADMIN_PASSWORD:-$RANDOM_ADMIN_PASS}
+ADMIN_PASSWORD=${RANDOM_ADMIN_PASS:-$ADMIN_PASSWORD}
```
swiftugandan, line 4: This would render ADMIN_PASSWORD redundant, remove it also remove it from Dockerfile

mgogoulos:
Did the change as you suggested @swiftugandan , tested and works as expected
Will update doc on https://github.com/mediacms-io/mediacms/blob/main/docs/Docker_Compose.md

Thanks!

PR #234, opened 2021-06-27
Added instructions inside robots_and_analytics.md

alberto98fx:
Added instructions to add Google Analytics and robots.txt

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/docs/robots_and_analytics.md
+++ b/docs/robots_and_analytics.md
@@ -1,0 +1,120 @@
+# Google Analytics
+
+1. Create a file:
+
+``` touch $DIR/mediacms/templates/tracking.html ```
+
+2. Add the Gtag/Analytics script
+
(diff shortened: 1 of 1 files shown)


mgogoulos:
This is great, thanks! 

Suggestion, what about adding the robots.txt as templates/cms/robots.txt and also on the code, and remove it from the instructions? We can merge this and I'd prefer we do so, since it is indeed something missing. 

Maybe also consider adding an entry on docs/Configuration.md ? Because this case definitely interests others. We'll move the documentation out at some point but until then it's good if it resides there. 

Regards
Markos

mgogoulos:
will add robots.txt and also the link to Configurations page. Thanks for the PR!

sinaqahremani:
Hi,
when I index all URLs of the site in Google Search Console, I get this error:

**Unknown video format**
What is your opinion? what can we do to solve this error?

PR #246, opened 2021-07-09
Framerate fix

multiflexi:
### 1 
Framerate should be kept unchanged if 60fps and less and as a fraction - as FFprobe reports it. Removed unwanted framerate change for videos with 60fps and less. Changed handling of higher framerates.

Sadly fails to encode if video is uploaded because of the "video_frame_rate": Fraction(video_info["r_frame_rate"]) bit for some reason, but float introduces unwanted rounding error.

### 2
A minimum bitrate modifier introduced as per https://developers.google.com/media/vp9/settings/vod


files/helpers.py
```diff
@@ -371,7 +373,7 @@ def media_file_info(input_file):
         "filename": input_file,
         "file_size": file_size,
         "video_duration": video_duration,
-        "video_frame_rate": float(Fraction(video_info["r_frame_rate"])),
+        "video_frame_rate": Fraction(video_info["r_frame_rate"]),
```
mgogoulos, line 376: This indeed makes files.helpers.media_file_info return something that is not JSON serializable and thus 

```
self.media_info = json.dumps(ret)
```

breaks on files.models.set_media_type

So if this is set to int, this won't break. But what is the problem you mention with cast to float?

slhck: Well, if you have a framerate that is actually a periodical number, this won't work and yield errors at latest when you try to re-encode specifying the (now rounded) framerate.

I think the most elegant way would be to store the framerate as numerator and denominator and use that, or keep the framerate as-is, except for cases where it's really really high.

files/helpers.py
```diff
@@ -501,10 +503,9 @@ def get_base_ffmpeg_command(
         enc_type {str} -- encoding type (twopass or crf)
     """
 
-    target_fps = int(target_fps)
-    # avoid Frame rate very high for a muxer not efficiently supporting it.
-    if target_fps > 90:
-        target_fps = 90
+    # avoid very high frame rates
+    while target_fps > 60:
+        target_fps = target_fps/2
```
slhck, line 508: What's the rationale for halving it? I guess it makes the pulldown easier.

files/helpers.py
```diff
@@ -687,8 +690,7 @@ def produce_ffmpeg_commands(media_file, media_info, resolution, codec, output_fi
     #        target_fps = 25
     #    else:
 
-    # adjust the target frame rate if the input is fractional
-    target_fps = src_framerate if isinstance(src_framerate, int) else math.ceil(src_framerate)
+    target_fps = src_framerate
```
slhck, line 693: This does not make sense if you override the `target_fps` in line 506, no? It will always be reset to the source framerate.

In principle I think keeping it exactly the same as the source framerate is meaningful provided the source framerate can be parsed correctly even if it's a fractional framerate.

The other alternative might be to not specify the framerate at all, and let ffmpeg encode it as-is. Just capture cases of too high framerates and scale them down.

files/helpers.py
multiflexi: Yes it is easier and you actually do not have to do a pull down of any sort, this just makes the encoded video to show every other, fourth, eight...frame, so no blending of frames is necessary. I set it to 60 as this is nowadays the maximum what displays can in most cases render and it is requires less CPU power to encode/decode. We can raise it to 120 in future when 120Hz displays become the mainstream.

files/helpers.py
multiflexi: Yes, I guess, I just did not wanted to break the naming of variables. The framerate should be kept the same as original except the case you want to provide <=30Hz encodes for videos with >30Hz (like YouTube does - and I think it does not make sense to have 240p60, but that should be left to user to decide in encoding profiles) and case when the framerate of original is >60Hz.

multiflexi:
Can this be merged or is there any issue?

mgogoulos:
> Can this be merged or is there any issue?

I'm going to test this out during the weekend and can merge it if nothing strange is found. 

Thanks and apologies for any delays

files/helpers.py
```diff
@@ -669,8 +686,8 @@ def produce_ffmpeg_commands(media_file, media_info, resolution, codec, output_fi
     else:
         return False
 
-    src_framerate = media_info.get("video_frame_rate", 30)
-    if src_framerate <= 30:
+    target_fps = Fraction(int(media_info.get("video_frame_rate_n", 30)), int(media_info.get("video_frame_rate_d", 1)))
```
slhck, line 689: Instead of defaulting to 30 here, I would rather let ffmpeg encode the input as-is. Fixing it to 30 could introduce unwanted frame duplications.

multiflexi:
If we leave it on FFmpeg to handle it then we cannot have 24/25/30Hz encodes of 48/50/60Hz source in future. I think this is good for low res where 60fps is not necessary. I forgot to test <1 fps source, will do it ASAP.

files/helpers.py
multiflexi: Yes, that is one option. The other is to think about the situation when this number is used. If I am not mistaken that happens only if the "video_frame_rate_n" value fails to be extracted by FFprobe. Then I guess the source is somehow broken or in a weird format, simply not standard. The question is what is better: trying to change the framerate to some standard value or letting FFmpeg decide. So basically I think it is frame duplication vs compatibility/non-deterministic result. I do not have a preference, either has its own benefits and downsides. Or am I wrong?

multiflexi:
OK, so I created a file with 1/5 fps (one frame every 5 seconds).

- Original does not play anywhere.
- In Firefox (Manjaro, Win10, macOS11) the minimum framerate which works is 8fps.
- In Chromium/Chrome (Manjaro, Win10, macOS11), Edge (Win10, macOS11) and Safari (macOS11) 1fps works fine.

So either the minimum will be 8fps because of how Firefox handles video or (what I prefer) the minimum will be 1fps and we raise an issue on their bug tracker because it works in other browsers.
Just for fun, I tested IE: the web is broken, but the video plays, the minimum is 5fps or somewhere between 1 and 5fps (but really who cares).

slhck:
> So either the minimum will be 8fps because of how Firefox handles video or (what I prefer) the minimum will be 1fps and we raise an issue on their bug tracker because it works in other browsers.
> Just for fun, I tested IE: the web is broken, but the video plays, the minimum is 5fps or somewhere between 1 and 5fps (but really who cares).

Interesting observations, thank you for testing! I think we can go for 1 just for safety measures, although I would prefer 8 just to make sure not to exclude Firefox users (which are a minority, but I don't think it's nice if Chrome were the only supported browser).

multiflexi:
Can we merge this or is there an issue? I already have another PR prepared. Or should I add it here?

files/helpers.py
```diff
@@ -366,18 +369,31 @@ def media_file_info(input_file):
...
+        "video_duration": video_duration, 
+        "video_frame_rate_n": video_frame_rate_n,
+        "video_frame_rate_d": video_frame_rate_d,
         "video_bitrate": video_bitrate,
         "video_width": video_info["width"],
         "video_height": video_info["height"],
         "video_codec": video_info["codec_name"],
         "has_video": has_video,
         "has_audio": has_audio,
+        "color_range": video_info["color_range"],
```
mgogoulos, line 390: I've uploaded a number of files and some of them (or all) break due to this. Retrieve more defensively all these variables , eg 

```
"color_space": video_info.get("color_space")
```

Are they going to be used somewhere?

files/helpers.py
```diff
@@ -366,18 +369,31 @@ def media_file_info(input_file):
         stdout = run_command(cmd).get("out")
         stream_size = sum([int(line) for line in stdout.split("\n") if line != ""])
         video_bitrate = round((stream_size * 8 / 1024.0) / video_duration, 2)
+    
+    if "r_frame_rate" in video_info.keys():
```
mgogoulos, line 373: can you output contents of `video_info["r_frame_rate"]` here? could this break now, or on future ffmpeg versions, so would it be better to ask it more defensively? After all these are not used anywhere (at least on this file!)

files/helpers.py
multiflexi: Yes, it will, for example for HDR->SDR conversion or conversion to Rec.709, but you are right it should be with get(). Should I fix it or can you do it?

files/helpers.py
multiflexi: I probably do not understand what you mean. Would you like it like this?
```
if video_info["r_frame_rate"]:
        video_frame_rate = video_info["r_frame_rate"].partition("/")
        video_frame_rate_n =  video_frame_rate[0]
        video_frame_rate_d =  video_frame_rate[2]  
```
It is used here:
```
target_fps = Fraction(int(media_info.get("video_frame_rate_n", 30)), int(media_info.get("video_frame_rate_d", 1)))
```
and there are default values for the case the if statement comes our false.

files/helpers.py
mgogoulos: since this is your PR, it's better if you fix them! thanks

files/helpers.py
mgogoulos: I was just worried whether there is a chance that video_frame_rate[0] or video_frame_rate[2] will result in IndexError!

files/helpers.py
multiflexi: Should be fixed now.

files/helpers.py
multiflexi: It is tuple. I tried to put nonsense in video_info["r_frame_rate"] but the worst case is that it does nothing. No IndexError.

mgogoulos:
This breaks due to pre-commit, I'm going to fix on a new PR now. 

Thanks a lot!!!

PR #251, opened 2021-07-21
[WIP] i18n changes

shubhank-saxena:


styiannis (changes requested):
1. Since you have not included in PR any change in file 'frontend/package.json', you should avoid including the changes of file 'frontend/package-lock.json'.
2. We should not replace texts derived from variables with static (translatable) texts. For example, it is valid to replace the text "Media Upload" with the its translatable form of "t('Media Upload')" (in React functional component using the appropriate hook) but it is not valid to replace the variable "PageStore.get('config-enabled').pages.members.title" with the translatable text "t('Members')", since in this way the procedure that sets the value to the variable is canceled.

PR #257, opened 2021-07-28
Feat urls

mgogoulos:
## Description
Adds three urls, documented on /swagger

* /api/v1/whoami
returns user info

* /'api/v1/user/token
returns user token

* 'api/v1/login
allows to login, through providing username or email and password


swiftugandan (commented):
Would be nice to have tests for new endpoints

mgogoulos:
> Would be nice to have tests for new endpoints

added tests for all 3 new endpoints - plus a new structure and a helper function to assist in adding new tests!

PR #259, opened 2021-07-31
install more recent ffmpeg

slhck:
## Description

Instead of using the system packages, install a recent ffmpeg static build.

This gets you more bugfixes and newer features.

Simplifies the OS check and package installation, because it was duplicated.

## Steps

*Pre-deploy*

*nothing*

*Post-deploy*

*nothing*


swiftugandan (commented):
The Docker files would also need to be updated

slhck:
Fixed, thanks for the hint.

mgogoulos:
I have tested this on a number of cases including Docker/server setups and it works without issues, merging

PR #264, opened 2021-08-02
move envs from Dockerfile to docker-compose files

mgogoulos:
## Description
Handle sensitive variables on docker-compose files rather than Dockerfile, thus changes are applied without having to build an image. 

We may optionally add env files to take care of them, but for now this fixes existing status where setting the password has no effect (same for the username)


mgogoulos:
@swiftugandan what do you think?

swiftugandan (commented):
@mgogoulos this looks ok

PR #265, opened 2021-08-03
Feat pdf viewer

swiftugandan:
## Description
* Adds PDF viewer using https://www.npmjs.com/package/pdf-viewer-reactjs
* Automatically builds the frontend and adds the static resources to the runtime docker image

TODO:
* For local deployment, we need to add a step that builds the frontend and copies the static resources to static
* If this is done, we shall not need to have the static resources in source control since we shall be generating them when the application is built.

@styiannis @mgogoulos what do you think?


styiannis:
@swiftugandan The library you have used to integrate the PDF viewer is quite good but we have to make some improvements in code before merge it into the main branch.

The main disadvantage is that the final size of the common css/js files that are used on every page is highly increased:

- static/js/common.js => from 430KB becames 1.1MB
- static/css/common.css => from 670KB becames 1.8MB

In order to limit the loading of PDF viewer files in the relevant page, we need first to separate (in django templates as well as in react components) the pdf viewer page from the rest of media pages (video, audio, etc..). 

Furthermore, we should make sure that by integrating the pdf viewer we will not load external sources that are already used by the system (for example, the package of ""material-design-icons"). 

Also, we should avoid loading CSS stylesheets and properties that are not necessary as well as check the appearance of the viewer in different screen sizes and make the appropriate improvements (in CSS) where it needed for better UX.

PR #273, opened 2021-08-12
CLI API Tool

shubhank-saxena:
## Description

This PR will create a CLI tool to interact with API


cli-tool/README.md
```diff
@@ -0,0 +1,10 @@
+## MediaCMS CLI Tool
+This is the CLI tool to interact with the API of your installation/instance of MediaCMS.
+
+### How to configure and use the tools
+- Make sure that you have all the required installations (`requirements.txt`)installed. To install it -
```
mgogoulos, line 5: maybe include the whole path here? `cli-tool/requirements.txt`

cli-tool/README.md
```diff
@@ -0,0 +1,10 @@
...
+    - Create a new virtualenv using any python virtualenv manager.
+    - Then activate the virtualenv and enter `pip -r requirements.txt`.
```
mgogoulos, line 7: `pip install -r `

cli-tool/cli.py
```diff
@@ -0,0 +1,168 @@
...
+    }
+
+    response = requests.post(url=f'{BASE_URL}/login', data=data)
+    if response.status_code == 200:
+        username = json.loads(response.text)["username"]
+        with open(".env", "w") as file:
+            file.writelines(f'AUTH_KEY={json.loads(response.text)["token"]}\n')
+            file.writelines(f'EMAIL={json.loads(response.text)["email"]}\n')
+            file.writelines(f'USERNAME={json.loads(response.text)["username"]}\n')
+        file.close()
```
mgogoulos, line 60: this is not necessary - file is closed automatically  when the with statement section finishes

cli-tool/cli.py
```diff
@@ -0,0 +1,168 @@
...
+@click.group()
+def apis():
+    """A CLI wrapper for the MediaCMS API endpoints."""
+
+
+@apis.command()
+def login():
+    """Login to your account."""
+
+    email = input('Enter your email address: ')
```
mgogoulos, line 45: email or username is valid, so this could ask for any of the two `enter email or username` type

cli-tool/cli.py
```diff
@@ -0,0 +1,168 @@
...
+    if response.status_code == 200:
+        username = json.loads(response.text)["username"]
+        with open(".env", "w") as file:
+            file.writelines(f'AUTH_KEY={json.loads(response.text)["token"]}\n')
+            file.writelines(f'EMAIL={json.loads(response.text)["email"]}\n')
+            file.writelines(f'USERNAME={json.loads(response.text)["username"]}\n')
+        file.close()
+        print(f"Welcome to MediaCMS [bold blue]{username}[/bold blue]. Your auth creds have been suceesfully stored in the .env file", ":v:")
+    else:
+        print(f'Error: {response.text}')
```
mgogoulos, line 63: this could be more friendly, rather than 
```
Error: {"non_field_errors":["User not found."]}
```
in case of error credentials, maybe print a message `check that credentials provided are correct, for URL`

cli-tool/cli.py
```diff
@@ -0,0 +1,168 @@
...
+
+
+def set_envs():
+    with open('.env', 'r') as file:
+        if not file.read(1):
+            print("Use the Login command to set your credential environment variables")
+        else:
+            global AUTH_KEY, USERNAME, EMAIL
+            AUTH_KEY = config('AUTH_KEY')
+            USERNAME = config('USERNAME')
```
mgogoulos, line 29: I've logged in, and see that .env creates the entry USERNAME
However here when I print USERNAME it is set to 'user' - AUTH_KEY and EMAIl are correct from .env

So my-media gives error results (queries `https://demo.mediacms.io/api/v1/media?author=user` instead of `https://demo.mediacms.io/api/v1/media?author=markos` in my case)

cli-tool/cli.py
shubhank-saxena: But if we want to keep both username or email, we will have to pass two different params in the request. And then two different variables/inputs would be needed from user

cli-tool/cli.py
shubhank-saxena: I am not able to replicate this error. I am getting correct media and username :(

mgogoulos:
Adding this as a baseline for the CMD client, thanks!

PR #348, opened 2021-12-19
Fix logo height

Megidd:
## Description

Trying to fix #107 

## Steps

As far as I tested, the problem is due to this statement:

```css
max-height: var(--logo-height, var(--default-logo-height));
```

Since neither `--logo-height` nor `--default-logo-height` is defined. I changed the `max-height` to a reasonable value. I used the font size of the root element which might be a very good base unit for the size of other UI elements:

https://stackoverflow.com/a/33028666/

# Test

As far as I test, the logo height looks good now.

### Before

Screenshot_20211219_233619

### After

Screenshot_20211219_233722


mgogoulos:
Thanks for opening this PR! Will test and see if it solves the error

Megidd:
Any update? :slightly_smiling_face:

mgogoulos:
> Any update? slightly_smiling_face

Not yet - sorry! Will prioritize this, thanks for your patience

mgogoulos:
just tested: it's better than the existing so I will merge it, however there's probably more fixes on this for another PR! Thanks

mgogoulos:
I've reverted these changes unfortunately, here's what I did

1. test these on an existing portal, by copying the two files, and setting another svg file. This worked well
2. once PR was merged, I created a new installation and saw that the default MediaCMS logo was showing very big...so I had to revert before we find out what is a more proper solution...

Megidd:
@mgogoulos Alright, no problem. Thanks for taking the time :slightly_smiling_face:

shodanx2:
Hello,
Just a confirmation that this is still an issue
I just tried placing a new banner and am getting
image
I will post my svg file for reference
I'm not sure if this can be openned without having the cookie monster font installed
LylyanneVannity-banner

I suspect the logo's aspect ratio is in cause ?

I tried making a two line text logo with a more square aspect ratio

This happenned
image

LylyanneVannity-banner two lines

Notice the text is higher than before, this is because instead of placing the text in the center of the canvas, I placed it at the top.

image

image


What this tells me is that there is "blank space" around my logo !

I thought the canvas was just a guide and not part of the actual .svg file !

Ok, I will now try making a landscape shaped canvas and making the logo the entire size of the canvas. 

So I made a new file and set document properties dimensions to 22.5 by 3.125 inch

And now I have much better results !


I made it fit as tight as possible, this is the best I was able to achieve

image

example files 
logo_dark
logo_light

mgogoulos:
Thanks for sharing, and great that you got it working! This PR was reverted because it introduced issues, so the issue with the logo is still there indeed. Hope your instructions help others until there is a more proper fix!

shodanx2:
Hi,
I have noticed my above attempt worked ok on a regular computer (although, as you can see there is still a lot of "blank" around the logo and so it is smaller than it should)

I noticed that on mobile the logo gets clipped for unknown reason (invisible element between logo and search magnifier icon ?)

image

BTW, here is how you change the canvas dimensions in Inkscape 

image

I created a new svg file with a narrow aspect ratio, this one is 13.5 horizontal by 6.5 vertical


LylyanneVannity-banner rectangle

Unfortunately this still clips on mobile

image

Also I notice the font I used didn't follow

And on PC the font is good but it's definitely too small to read

image

Ok,  for the font I found this answer

https://graphicdesign.stackexchange.com/questions/27238/preserving-font-type-in-illustrator-svg-file

The trick is to convert the font into an outline

It is in the menu Path -> Object to path

image

(This made my 2kb svg file into a 23kb svg file, still acceptable)

I also moved the second row text a little bit to the left to avoid clipping

Current svg file


LylyanneVannity-banner rectangle object to path

Now it looks like this on my pc

image


and like this on my phone

image

If I zoom in very far, we can see the font is intact and the whole thing is not clipped, however it is too small to read

zoomed in on PC

image

zoomed in on phone

image


However for now this is as good as I can make it as I don't know much of anything on html ......


Although, this is the relevant section in the html code, viewing with firefox F12 html inspector console

image

And I have made a small video using the object inspector to show what element does what

I think the following code is what an html expert should be looking at to fix this issue

```
<div class="page-header-left">
<div>
<div class="close-search-field">
<button class="circle-icon-button">
<span><span>
<i class="material-icons">arrow_back</i>
</span></span>
</button></div>
<div class="toggle-sidebar">
<button class="circle-icon-button">
<span><span>
<i class="material-icons">menu</i>
</span></span></button>
</div><div class="logo">
<a href="/" title="Lylyanne Vannity">
<span><img src="/static/images/logo_light.svg" alt="Lylyanne Vannity" title="Lylyanne Vannity" loading="lazy">
</span></a></div></div></div>
<div class="search-field-wrap"><div>
<form method="get" action="/search" autocomplete="off">
<div><div class="text-field-wrap"><input type="text" placeholder="Search" aria-label="Search" name="q" value=""></div>
<button type="submit" aria-label="Search"><i class="material-icons" data-icon="search">
</i></button></div></form></div></div><div class="page-header-right"><div>
<div class="mobile-search-toggle"><button class="circle-icon-button" aria-label="Search">
<span><span><i class="material-icons" data-icon="search"></i></span></span></button>
</div><div class="user-thumb"><button class="circle-icon-button thumbnail small-thumb" aria-label="Account profile photo that opens list of options and settings pages links">
<span><span><img src="/media/userlogos/2022/10/30/lyly-souris.jpg" alt=""></span></span></button></div></div></div>
```

bananacy:
Still unable to replace logo, its invisible. but there in the source code

shodanx2:
Try making a test logo file in inkscape first maybe ? 
The structure of the svg file seems to be important.
With inkscape, make an svp file with just the letters "ABC"in the middle and try that. 
Sometimes I had no logo, but it's because the logo was outside the display space, there was too much white space on the left of the logo

beatboxa:
Did anyone ever get this to work; and if so, can anyone provide step-by-step instructions?

I have tried changing the logo-height, default-logo-height, etc. it seems just about everywhere.
By default, it appears to be 18px.

If I right click the logo in a browser and hit "Inspect Element,"
And then I search for "logo-height" under Styles on the very right;
I come across this:
Screenshot 2025-01-29 at 10 33 57 PM

Changing that "--default-logo-height" to "36px" here temporarily makes my logo into the size that I want.  The issue is that I don't know how to trace this back to the code; and even if I think I can, I don't know how to apply the changes.

I came across this:  https://github.com/mediacms-io/mediacms/issues/541

But I am running via docker; and I don't know if/how I can run that because npm is not installed.

Changing a logo and sizing it appropriately seems like one of the most common use cases; but this seems like quite a lot of effort to do something that should be so simple.

mfkp:
We now have the ability to update a new logo easily as per this recent update: https://github.com/mediacms-io/mediacms/pull/1391

Unfortunately it still defaults to 18px logo height with SVG logos, and this is very tough to fix on a docker deployment... I'm still investigating to see if there's a hopefully easy fix for this.

mfkp:
If we change that max-height attribute to 100%, it works just fine. I think that whoever is providing a custom logo should make sure that it looks good in the header within the confines of the bounding box (aka providing a horizontal logo and not a vertical logo). Thoughts @mgogoulos ?

## Before
18px height logo

## After
100% height logo

Issue #123, opened 2021-03-30
[BUG] Like/Dislike button malfunction 


JohnPapad:
Hello, when the 'Like' button is pressed, it can not be undone (by pressing it again) or changed to 'Dislike' and vice versa. 
It's allowed for a video to be classed as both Liked and Disliked, at the same time.
In case the page is reloaded, a previously Liked video does not appear as Liked anymore and the user can press the 'Like' button again, resulting to the corresponding counter's increase (same goes for the 'Dislike' button). 
The 'Likes/Dislikes' bar does not show the correct ratio.

Megidd:
At first, the click on like button works fine:

![image](https://user-images.githubusercontent.com/17475482/146628754-53f6d23e-7105-41c4-b0d3-f3a68b523b4c.png)

But then, visiting the page again, the like/dislike count is just the same. Looks like they are bound together:

![image](https://user-images.githubusercontent.com/17475482/146628780-c97df934-44c9-4369-81c0-d2fec47a3fe0.png)

![image](https://user-images.githubusercontent.com/17475482/146628579-b34265c7-0687-45d8-b74d-d5ddf0ab208e.png)

Megidd:
Source file `frontend/src/static/js/components/media-actions/MediaDislikeIcon.jsx` has a statement:

```js
MediaPageActions[dislikedMedia ? 'undislikeMedia' : 'dislikeMedia']();
```

Source file `frontend/src/static/js/utils/actions/MediaPageActions.js` has a `dislikeMedia` function. But it doesn't have any `undislikeMedia` function.

Maybe that's related to the like/dislike bug.

mgogoulos:
this is now fixed

PR #388, opened 2022-01-09
Fix dislike count

Megidd:
Trying to fix dislike counter bug reported here: https://github.com/mediacms-io/mediacms/issues/123#issuecomment-997141350


Megidd:
# Test

Running these on my development machine shows the dislike counter is now fixed:

```bash
docker-compose -f docker-compose-dev.yaml exec -T frontend npm run dist
cp -r frontend/dist/static/* static/
```

_Dislike_ counter is **not** equal to _like_ counter any longer:

Screenshot_20220110_185118

PR #429, opened 2022-03-15
Mention users in comments

MrPercheul:
## Description
This change adds the react component "react-mentions" (https://github.com/signavio/react-mentions) in the comment section so that users can tag one another.

When typing '@' in a comment, a suggestion list will be rendered will all the users, and you can pick one to add him to the comment. The name will be highlighted following the color scheme (10% lighter/darker than #009933) and saved on the BD as "texttext **@[_IDUser_](_DisplayName_)** texttext"

Comments rendered under the media will watch for these markups and replace it with the appropriate anchor (<a href="/user/**IDUser**>**DisplayName**</a>)


files/methods.py
```diff
@@ -347,6 +349,58 @@ def notify_user_on_comment(friendly_token):
...
+    for match in list(dict.fromkeys(matches)):
+        notify_user_on_mention(friendly_token, match, cleaned_comment)
+
+
+def clean_comment(raw_comment):
+    """ Clean the comment fromn ID and username Mentions for preview purposes """
+
+    cleaned_comment = re.sub('@\\(_(.+?)_\\)', '', raw_comment)
+    # cleaned_comment = "<div>".join(cleaned_comment.split("[_"))
+    # cleaned_comment = "</div>".join(cleaned_comment.split("_]"))
```
swiftugandan, line 397: delete?

files/methods.py
```diff
@@ -347,6 +349,56 @@ def notify_user_on_comment(friendly_token):
     return True
 
 
+def notify_user_on_mention(friendly_token, user_mentionned, cleaned_comment):
```
mgogoulos, line 352: s/user_mentionned/user_mentioned

files/methods.py
```diff
@@ -347,6 +349,56 @@ def notify_user_on_comment(friendly_token):
...
+
+    """ Notify the user mentionned in the comment """
```
mgogoulos, line 354: s/mentionned/mentioned

files/methods.py
```diff
@@ -347,6 +349,56 @@ def notify_user_on_comment(friendly_token):
...
+    media = None
```
mgogoulos, line 355: this line is not needed

files/methods.py
```diff
@@ -347,6 +349,56 @@ def notify_user_on_comment(friendly_token):
...
+            media_url,
+            cleaned_comment,
+        )
+        email = EmailMessage(title, msg, settings.DEFAULT_FROM_EMAIL, [user.email])
+        email.send(fail_silently=True)
+    return True
+
+
+def check_comment_for_mention(friendly_token, comment_text):
+    """ Check the comment for any mentions, and notify each mentionned users"""
```
mgogoulos, line 381: s/mentionned/mentioned

mgogoulos (changes requested):
First of all sync with master to solve the conflict (after I merged your other PR!) 

Besides some minor issues added here (mostly typos), the biggest issue with it is that it adds a call to users API every time you visit a media page. This cannot be the behavior out of the box, so I suggest that you make it configurable, and that by default this feature is disabled. 

For example it can be a setting on cms/settings.py as MENTION_USERS_IN_COMMENTS that is False (and on local_settings.py you override with True). Then you may pass this variable on files/context_processors.py, and have the frontend read it as it does on  templates/config/  ...

MrPercheul:
The changes are in and the Typos corrected, however I did not get a precommit black failed when running it on my side, anyway of checking which file is the cultprit?

files/methods.py
MrPercheul: typo fixed

files/methods.py
MrPercheul: Typo fixed

files/methods.py
MrPercheul: Typo fixed

files/methods.py
MrPercheul: Typo fixed

MrPercheul:
There is an issue with Pre-commit, the 'click' library used by black is causing trouble, it is fixed in this commit and I will also submit another PR just for the .pre-commit-config.yaml change

files/methods.py
```diff
@@ -347,6 +349,54 @@ def notify_user_on_comment(friendly_token):
     return True
 
 
+def notify_user_on_mention(friendly_token, user_mentioned, cleaned_comment):
+    media = None
```
mgogoulos, line 353: I am refering to the `media = None` here. 

media on this queryset will be None if it doesn't find one, that's why I commented that `media = None` is redundant!
```
media = models.Media.objects.filter(friendly_token=friendly_token).first()
```

files/methods.py
```diff
@@ -347,6 +349,54 @@ def notify_user_on_comment(friendly_token):
...
+    media = models.Media.objects.filter(friendly_token=friendly_token).first()
+    if not media:
+        return False
+
```
mgogoulos, line 357: can you bring the `from users.models import User` here, and not on the top level? Just in case no weird cyclic import error can  happen

mgogoulos:
looks good, I will give it another test and merge it once you make the following two changes:

1. handle last 2 small issues
2. add an entry on the documentation, on how this is used, and probably a screenshot too

Thanks!

MrPercheul:
Hi, the changes are in, and the entry in the user_docs.md too

mgogoulos:
merging this, very useful functionality, thanks for introducing!

PR #430, opened 2022-03-17
Timestamp

MrPercheul:
Small change to be able to link and start videos at a certain timestamp a la youtube
-> Can start a video a time X by adding the GET param 't' in the URL
-> The share button as a 'start at XX:XX' option
-> Comments including a timestamp (HH:MM:SS or shorter) are wrapped in a anchor tag


mgogoulos:
Hi, had a look on it and looks great! Some notes, before we can merge:

1. Add a check that parameter 't' is indeed an integer, otherwise ignore it. So that if you provide http://url/view?m=EmF7XfDIE&t=xyz it gets ignored and doesn't raise an error

2. Then have 't' compared to the duration of the video (which is available at that point). If t is >= duration, then set t as the duration.

These two don't create serious issues if they are misused but I believe are easy to handle so we'd better do!

3. Add an entry on the documentation for both of these (ability to provide timestamp, plus the addition on the comments). Probably this page is suitable for it:  https://github.com/mediacms-io/mediacms/blob/main/docs/user_docs.md
You may include a mention how you can use it, even if you want to provide a screensot as well.


Thanks

mgogoulos:
Thanks @MrPercheul ! I will merge this, and then create a new PR (with some typo fixes on the doc, plus the static files, plus a mention on the authors/contributors file)

PR #552, opened 2022-10-20
Wrap logo in h1 tag for search engines and screenreaders

KyleMaas:
## Description
This adds an `<h1>` tag around the logo in the header, mainly to help search engines and screenreaders find the name of the site, since the rest of the page was missing header information.

## Steps

*Pre-deploy*
```
cd frontend
npm run dist
cp -ra dist/static/* ../static/
```

*Post-deploy*

### DIFF
--- a/frontend/src/static/js/components/page-layout/PageHeader/Logo.jsx
+++ b/frontend/src/static/js/components/page-layout/PageHeader/Logo.jsx
@@ -4,9 +4,11 @@ export const Logo = ({ src, loading = 'lazy', title, alt, href = '#' }) => {
   return src ? (
     <div className="logo">
       <a href={href} title={title}>
-        <span>
-          <img src={src} alt={alt || title} title={title} loading={loading} />
-        </span>
+	<h1>
+          <span>
+            <img src={src} alt={alt || title} title={title} loading={loading} />
+          </span>
+	</h1>
       </a>
     </div>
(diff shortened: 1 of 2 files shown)


KyleMaas:
Also added static indexable links for the nav menu.

templates/components/sidebar.html
```diff
@@ -1 +1,26 @@
-<div id="app-sidebar"></div>
\ No newline at end of file
+<div id="app-sidebar">
```
mgogoulos, line 1: Why are you adding these at that place? I think it's better to avoid bringing all of them here for a number of reason (they might be removed by some installations, or even the URLs changed etc)

KyleMaas: Mainly because that's where they are generated in the frontend, so when they're replaced during load, the markup is similar.  This means it's compatible with the existing stylesheet to prevent it from blinking and rearranging as the page progressively loads on slower connections.  The differences from the final frontend-generated version also mean that any search engines indexing the page which do partially evaluate JavaScript (such as Google) don't ding you for presenting something different to users than what the search engine sees.

mgogoulos:
Thanks for this work @KyleMaas , I'm closing down as we need to refactor the way all basic components are loaded, get out of React and load from single Django html template. This is planned for the next 1-2 months.

PR #557, opened 2022-10-22
Add static content for media pages

KyleMaas:
## Description
Media pages are essentially unsearchable right now because all of the relevant information is contained within JavaScript.  This change adds some of the metadata for media as static textual content on the page so it can be properly indexed.

## Steps

*Pre-deploy*

*Post-deploy*


KyleMaas:
Made a few additional changes to include subtitle content as searchable text as well and to hide the metadata during page load.

KyleMaas:
I should note that the changes that I made to hide metadata during page load require a frontend rebuild, but I wasn't sure how best to go about submitting that.  Committing the built files seems like a great way to cause conflicts, so I did not do that.

mgogoulos:
Thanks for this work! I see the point but I'm very skeptical for a number of reasons.

This is adding lots of extra queries for the index/media pages, and that could impact a heavy portal. 
These queries could be optimized to get only the necessary information (eg use Django tools as the 'only' on a queryset to get the list of items -title/description- and not the whole objects. 

However isn't there redundancy in making these queries twice (one to get the metadata, and one to get it through the API). 

Keep in mind that if the template takes time to load (because it's making these queries), then the page won't show up anything (because the template has to load, and then the frontend is making the API queries). 

If media data are returned to the Django template, then what is the reason to call the API - all data could be returned to the template, and this would give more consistent results. 

What do you think?

KyleMaas:
Potentially, yes, but that seemed like a much deeper architectural change than I'd want to do without discussion.  This solution, while it increases the initial page load time slightly, *significantly* improved searchability.  Having the metadata embedded into the page as JSON and eliminating the API call would be a great way to have both the searchability, better compatibility with non-JS browsers and screen readers, and eliminate the duplication.

Thank you for your feedback on this.

mgogoulos:
Closing this as it is old, my bad for not reviewing in time. @KyleMaas please bring these ideas once we complete the refactoring where Django templates replace what is loaded from React now

PR #558, opened 2022-10-22
Fix double slashes in URIs

KyleMaas:
## Description
This fixes an issue where URIs all over the system were output in the form:

```
https://example.com//view?m=(video ID)
```

...even when the `FRONTEND_HOST` and `SSL_FRONTEND_HOST` did not have a trailing slash.  This was especially problematic with the meta `canonical` tag, which was pretty much always a different URI than the page that was outputting it, but many of the other URIs on the page (especially in the metadata) were including the double slashes as well.

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/files/context_processors.py
+++ b/files/context_processors.py
@@ -6,7 +6,7 @@ from .methods import is_mediacms_editor, is_mediacms_manager
 def stuff(request):
     """Pass settings to the frontend"""
     ret = {}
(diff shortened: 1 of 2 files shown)


mgogoulos:
great catch! I did not know this extra slash was creating other problems, other than being not nice to see. 
I will merge this now, tested it locally and it's working. I would feel more confident if there were more tests in place but what can you do!

PR #559, opened 2022-10-22
Truncate titles and descriptions when they exceed limits

KyleMaas:
## Description
Users on an instance I'm running like to use really long titles and descriptions.  Per:

https://www.w3.org/Provider/Style/TITLE.html#:~:text=The%20title%20should%20ideally%20be,there%20is%20only%20limited%20room.

...titles should only be 64 characters long, and meta descriptions should usually be limited to 160 characters.  This change truncates both if they exceed limits to keep things in compliance.

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/templates/cms/media.html
+++ b/templates/cms/media.html
@@ -11,7 +11,7 @@
 
 <meta property="og:title" content="{{media_object.title}} - {{PORTAL_NAME}}">
 <meta property="og:url" content="{{FRONTEND_HOST}}{{media_object.get_absolute_url}}">
-<meta property="og:description" content="{% if media_object.summary %}{{media_object.summary}}{% else %}{{media_object.description}}{% endif %}">
+<meta property="og:description" content="{% if media_object.summary %}{{media_object.summary|truncatechars:300}}{% else %}{{media_object.description|truncatechars:300}}{% endif %}">
 <meta property="og:updated_time" content="{{media_object.edit_date}}">
 
 {% if media_object.media_type == "video" %}
--- a/templates/cms/media.html
+++ b/templates/cms/media.html
@@ -1,13 +1,13 @@
 {% extends "base.html" %}
 {% load static %}
 
-{% block headtitle %}{{media_object.title}} - {{PORTAL_NAME}}{% endblock headtitle %}
+{% block headtitle %}{{media_object.title|add:" - "|add:PORTAL_NAME|truncatechars:64}}{% endblock headtitle %}
 
 {% block headermeta %}
 
 <link rel="canonical" href="{{FRONTEND_HOST}}{{media_object.get_absolute_url}}">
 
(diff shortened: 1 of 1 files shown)


templates/cms/media.html
```diff
@@ -1,13 +1,13 @@
 {% extends "base.html" %}
 {% load static %}
 
-{% block headtitle %}{{media_object.title}} - {{PORTAL_NAME}}{% endblock headtitle %}
+{% block headtitle %}{{media_object.title|add:" - "|add:PORTAL_NAME|truncatechars:64}}{% endblock headtitle %}
```
mgogoulos, line 4: I think 
`{{media_object.title|truncatechars:5}} - {{PORTAL_NAME}}` is what you want here, if you want to truncate the title and keep the portal name at the end

templates/cms/media.html
```diff
@@ -1,13 +1,13 @@
...
 
-{% block headtitle %}{{media_object.title}} - {{PORTAL_NAME}}{% endblock headtitle %}
+{% block headtitle %}{{media_object.title|add:" - "|add:PORTAL_NAME|truncatechars:64}}{% endblock headtitle %}
 
 {% block headermeta %}
 
 <link rel="canonical" href="{{FRONTEND_HOST}}{{media_object.get_absolute_url}}">
 
-<meta name="description" content="{% if media_object.summary %}{{media_object.summary}}{% else %}{{media_object.description}}{% endif %}">
+<meta name="description" content="{% if media_object.summary %}{{media_object.summary|truncatechars:160}}{% else %}{{media_object.description|truncatechars:160}}{% endif %}">
```
mgogoulos, line 10: consider doing the same for the `meta property="og:description"` too ?

mgogoulos:
This is unrelated with the scope of this PR, however if you have the capacity to add it it would be great to wrap the description below the media (with a show more option) - since now it will display the whole description

KyleMaas:
Fair enough.  I'll see what I can do.

templates/cms/media.html
KyleMaas: In this case, I figured it was more important to have the media title in there than to always have the portal name in there, which is why I truncated the concatenation instead.

templates/cms/media.html
KyleMaas: Added as a commit with a limit of 300, since there doesn't seem to be a hard limit on that like there is on meta description tags but 300 seems to be a commonly-agreed-upon maximum length for that.

KyleMaas:
> This is unrelated with the scope of this PR, however if you have the capacity to add it it would be great to wrap the description below the media (with a show more option) - since now it will display the whole description

I split this off into its own issue (#660) so I can remember to work on it.

PR #572, opened 2022-11-14
Add sitemap

KyleMaas:
## Description
Adds a basic sitemap XML file for the site which can be used for crawling and indexing it.

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
This should not be enabled by default, since the database queries on files/views.py could be very problematic for large portals - eg hosting 10k's of media or more. For smaller installations that could be fine

May I suggest you add two things here:
1. a setting on settings.py (eg GENERATE_SITEMAP) that defaults to False, and based on it the sitemap function on files/views.py either returns False or does the work
2. add an entry on docs/  probably on the admin document, on how to enable sitemap?

Thanks

KyleMaas:
Sure, I could do that.

KyleMaas:
That work?

KyleMaas:
Thanks!

PR #677, opened 2022-12-18
Cross-browser event path

Micrufun:
Fix errors on web browsers due to usage of the  non-standard `path` property of event objects:

Firefox:

Screenshot_20221218_181237--crop

Chromium:

Screenshot_20221218_181541--crop


Micrufun:
Sorry. It was the wrong branch  I was trying to merge.

PR #682, opened 2022-12-26
Encode profiles doc

Micrufun:
It's clarified how to exactly modify `encode profiles`. Looks like the previous statements were wrong.


mgogoulos:
thanks for opening!

Issue #694, opened 2023-01-12
TIMESTAMP_IN_TIMEBAR does not work.


jp-rad:
**Describe the issue**
A clear and concise description of what the issue is.

**To Reproduce**

1. TIMESTAMP_IN_TIMEBAR = True  # shows timestamped comments in the timebar for videos
1. Add a comment, 'mm:ss` or `hh:mm:ss`
1. Comment without timestamped

**Expected behavior**
1. shows timestamped comments in the timebar for videos

jp-96:
I was mistaken.
The timestamp on comments is a specification that always works. But it doesn't seem to work on the loadedmetadata event. canplay event and it works as expected.
But the markers are unstable.

mgogoulos:
it seems that timestamp has broken after video.js upgrade - https://github.com/mediacms-io/mediacms/pull/519

trying to fix, any help is appreciated!

jp-rad:
I will try.

mgogoulos:
To give more insight here, the video.js upgrade on https://github.com/mediacms-io/mediacms/pull/519 made this feature break, and it's strange how a video.js (an external library) could trigger this issue. When you specify timestamp on the URL and start a video, you see that it starts on the correct time but immediately after that it starts from the beginning of the video, as sure as a function is triggered that does this

jp-rad:
`setTimestampAnchorsAndMarkers()` and `initialize()` are called on the `loadedmetadata` event, and thus the order of their execution is undefined.

`setTimestampAnchorsAndMarkers()`
https://github.com/mediacms-io/mediacms/blob/main/frontend/src/static/js/components/comments/Comments.jsx#L447

`initialize()`
https://github.com/mediacms-io/mediacms/blob/main/frontend/src/static/js/components/comments/videojs-markers.js#L454

jp-rad:
`setTimestampAnchorsAndMarkers()` is only called on the `loadedmetadata` event, so adding or removing comments during video playback will not change the timebar markers.

https://github.com/mediacms-io/mediacms/blob/main/frontend/src/static/js/components/comments/Comments.jsx#L447

mgogoulos:
I think the problem is related with the video player doing something wrong, so not necessarily related with comments, I guess it's something on frontend/src/static/js/components/video-player/VideoPlayer.jsx

jp-rad:
Pull Requested, https://github.com/mediacms-io/mediacms/pull/706
[...]

PR #706, opened 2023-01-30
Fixed event handler for displaying comments and adding markers to the timeline

jp-rad:
Fixed event handler for displaying comments and adding markers to the timeline

https://github.com/mediacms-io/mediacms/issues/694

## Description

Changing `loadedmetadata` event to `loadeddata` event, because called after `initialize()`.

`initialize()`
https://github.com/mediacms-io/mediacms/blob/main/frontend/src/static/js/components/comments/videojs-markers.js#L454

The original `onCommentsLoad` function was separated into the `onCommentsLoad` function and the `retrieveComments` function.

### DIFF
--- a/frontend/src/static/js/components/comments/Comments.jsx
+++ b/frontend/src/static/js/components/comments/Comments.jsx
@@ -422,37 +422,51 @@ const CommentsListHeader = ({ commentsLength }) => {
 export default function CommentsList(props) {
   const [mediaId, setMediaId] = useState(MediaPageStore.get('media-id'));
 
-  const [comments, setComments] = useState(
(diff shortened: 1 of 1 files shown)


mgogoulos:
Hi @jp-rad 
I've checked out this PR and have the following questions:

1. Is this ready to be merged or you have concerns? Not sure I understand what is the case on https://github.com/mediacms-io/mediacms/issues/694#issuecomment-1408584148
2. The big problem remains after update to latest video.js and is the fact that when you visit a media URL that has timestamp, the video starts on this timestamp but IMMEDIATELY afterwards goes back to the beginning. This behaviour wasn't happening before. Do you have any idea how this can be fixed?

mgogoulos:
this is now fixed

PR #749, opened 2023-03-27
Upgrade postgres docker compose

Adito5393:
## Description
This PR includes the following changes:
- Upgraded the DB to the latest stable alpine version. The version is pinned with MAJOR.MINOR (15.2), but can be changed to MAJOR (15) only if desired.
- Updated the health check command to use environment variables instead of a hard-coded value.
- Added the timezone environment variable TZ to the DB.

## Testing
The changes were tested on a VM server using the docker compose stacks.

### DIFF
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -40,8 +40,9 @@ services:
       POSTGRES_USER: mediacms
       POSTGRES_PASSWORD: mediacms
       POSTGRES_DB: mediacms
+      TZ: Europe/London
     healthcheck:
-      test: ["CMD-SHELL", "pg_isready -U mediacms"]
+      test: ["CMD-SHELL", "pg_isready", "--host=db", "--dbname=$POSTGRES_DB", "--username=$POSTGRES_USER"]
       interval: 10s
       timeout: 5s
       retries: 5
--- a/docker-compose-dev.yaml
+++ b/docker-compose-dev.yaml
@@ -32,7 +32,7 @@ services:
       db:
         condition: service_healthy
   db:
-    image: postgres:13
+    image: postgres:15.2-alpine
     volumes:
       - ../postgres_data:/var/lib/postgresql/data/
     restart: always
--- a/docker-compose-http-proxy.yaml
+++ b/docker-compose-http-proxy.yaml
@@ -68,7 +68,7 @@ services:
     depends_on:
       - migrations
   db:
-    image: postgres:13
+    image: postgres:15.2-alpine
     volumes:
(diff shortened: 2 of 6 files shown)


mgogoulos:
Looks good, however I'm wondering whether it needs anything for installations that are currently using PG 13. Do they require some sort of migration, or this is happening transparently between versions 13 and 15?

mgogoulos:
I just tried that on an installation that was having MediaCMS running through Docker, with PG 13. It does not run, and the message is this: 

```
db_1              | PostgreSQL Database directory appears to contain a database; Skipping initialization
db_1              | 
db_1              | 2023-06-27 11:07:42.959 UTC [1] FATAL:  database files are incompatible with server
db_1              | 2023-06-27 11:07:42.959 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 13, which is not compatible with this version 15.2.
```

So although new installations will work fine with this, it cannot be merged since it will break existing installations ;(

mgogoulos:
If this is something that cannot be fixed automatically, at least a complete migration path should be included, with ready to run instructions for exporting/deleting/importing data...

Adito5393:
Indeed, running the DB inside the docker container does not mean that [pgupgrade](https://www.postgresql.org/docs/current/pgupgrade.html) is useless. All users who self-host services, sooner or later learn how to migrate DB depending on which DB they favor.

There are lots of guides on PostgreSQL:
- if downtime is not an issue, via dump & restore: 
  - https://eelkevdbos.medium.com/upgrade-postgresql-with-docker-compose-99d995e464
  - https://betterprogramming.pub/how-to-upgrade-your-postgresql-version-using-docker-d1e81dbbbdf9?gi=c1d6a7be7b41
  - If `POSTGRES_PASSWORD_FILE` was used, changing the encryption might be required: https://christierney.com/2022/03/09/upgrading-postgresql-on-docker-compose/
  - https://thomasbandt.com/postgres-docker-major-version-upgrade
- downtime is an issue, then a container with both major versions installed can be used to trigger the `pgupgrade`, see the PoC [tianon/docker-postgres-upgrade](https://github.com/tianon/docker-postgres-upgrade) repo.

I think MediaCMS should not recommend users to blindly use the available `docker-compose` files but offer them as examples on how to initially setup. Most people who use docker will not even `git clone` the repo & just handy pick the file(s) needed.
I only needed to overwrite the `deploy/docker/local_settings.py`

mgogoulos:
Merging, will now add notes on the docs on how to update - https://github.com/mediacms-io/mediacms/blob/main/docs/admins_docs.md

PR #761, opened 2023-04-19
ansible playbook 

zaterio:
## Description

This PR implements ansible playbook for deploying mediacms on remote hosts, as an alternative to install.sh script.

PR Description:

1. A directory called "ansible" has been created in the root of the repository.
2. No other additional changes have been made outside of the "ansible" directory
3. In the "ansible" directory the starting point of the application is the "hosts" file, which is where the different deploy hosts are declared.

hosts file operation:

- in the hosts file, 3 groups can be declared: **mediacms_servers**, **postgresql_servers** and **redis_servers**.
- **postgresql_servers** group: one host can be declared, In this case, postgres installation will not be performed on the mediacms server. In case there are more than one, only the first one will be selected. If the **postgresql_servers** group is not declared, the installation will be performed on localhost and the behavior is the same as install.sh. Only Debian bullseye postgresql server allowed for remote hosts (This is due to the presence of the postgres system user, by default in this distro, which is used to create the database user and migrations). Remote postgresql servers must have python3-psycopg2 package installed and listen on default port 5432.  playbooks does not take care of installation on remote postgres servers. the playbook takes care of configuring DATABASES appropriately in settings.py, independent if it is on localhost on remote host.

- **mediacms_servers** group: only one host can be declared. In case more than one are added, redis_servers must not be declared, this is because there would be collisions in the celery queues.
-  **redis_servers** group: only one host can be declared, in case the redis_servers group is not added, the installation of redis-server will be done on localhost, equal to install.sh.

In short, hosts file should be declared as:

```
[mediacms_servers]
testmediacms1 ansible_port=22

[postgresql_servers]
postgresql ansible_port=22

[redis_servers]
redis ansible_port=22
```

with this it is possible to separate redis, postgres and mediacms.

or:
```
[mediacms_servers]
testmediacms1 ansible_port=22
```
in this way the same installation is produced as in install.sh

Assuming you are located inside the ansible directory, and you have edited the hosts file,  and have ansible installed on your computer, and you have ssh access to the remote machines mediacms... etc. mediacms is installed on the remote machine by running:

`ansible-playbook playbooks/01_system_deploy.yml`

Regards.

### DIFF
--- a/ansible/ansible.cfg
+++ b/ansible/ansible.cfg
@@ -1,0 +1,11 @@
+[defaults]
+inventory  = hosts
+host_key_checking = False
+deprecation_warnings = False
+roles_path = roles
+
+[ssh_connection]
+# ssh arguments to use
+# Leaving off ControlPersist will result in poor performance, so use
+# paramiko on older platforms rather than removing it, -C controls compression use
+ssh_args = -C -o ControlMaster=auto -o ControlPersist=60s -o ForwardAgent=yes
--- a/ansible/hosts
+++ b/ansible/hosts
@@ -1,0 +1,17 @@
+# only one media server can be added, at the moment there is no celery queue separation at the redis level, which implies coding errors
+
+[mediacms_servers]
+testmediacms1 ansible_port=22
+
+# If postgresql_servers is declared, it must contain a single host.
+# Only Debian bullseye postgresql server allowed.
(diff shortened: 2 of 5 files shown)


mgogoulos:
Hi @zaterio , thank you very much for opening the PR. I didn't find the time yet to test it before I can merge it, and now I see you have closed it. If you think it is still valid feel free to reopen! 

Thanks and regards

PR #763, opened 2023-04-24
Integration of Owncast with MediaCMS for live streaming

KyleMaas:
## Description
Integration of Owncast with MediaCMS.  After running this now for several months, there's more I'd like to add but I feel it's stable enough for primetime.

Setup for this is really simple.  This operates using two new settings which can be added to `local_settings.py`:

`LIVESTREAM_BACKEND` -> Set to "owncast" to enable the live stream functionality.  This leaves open the possibility (which I intend to use in the future) of tying into alternative live streaming platforms.
`LIVESTREAM_URI` -> Set to the base address of your Owncast installation, e.g. "https://live.example.com/"

When enabled, it adds an item to the left nav bar called "Live", and if it's able to successfully connect to Owncast and query its status, it will change to "Live (Offline)" or "Live (Online)" depending on the very resource-minimal Owncast status API, which then gets queried once per minute from the client end so people can watch right away when a stream goes live.  If the stream is live when the homepage is loaded, it also shows the live stream at the top with a thumbnail and count of current viewers, again using the information queried from Owncast.

Eventually I'd love to have the left nav bar item open in a new window.  But that's really about the only thing I'm not 100% satisfied with about the way this has been working for me.

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
Thanks a lot for the PR! Would it be possible to share a screenshot, or even include a section on the documentation?

KyleMaas:
Not sure how to resolve that conflict, but I added documentation as to how to use this.  Due to the project this is being used on, I cannot include a screenshot.  Hopefully that's enough to explain how it works.  Feel free to change as needed to resolve the conflict.

morbificagent:
Hi together,
i love the idea of being able to integrate owncast here...
Im not very common in Github pull request coding things ;-) but why isnt this implemented yet, as for me as a "no-coder" it looks like a complete code from @KyleMaas which could be used for this...

morbificagent:
OK, have tried to make the same changes as you showed in your 3 commits above but nothing has changed on my site... Do i have to do something more as make the same changes? 

Many thanks for your help!

KyleMaas:
@morbificagent 

Awesome!  Glad it's working for you.  And now we have a screenshot to work with.

KyleMaas:
Attempted fixing merge conflicts

files/context_processors.py
```diff
@@ -31,4 +31,6 @@ def stuff(request):
     ret["ALLOW_RATINGS_CONFIRMED_EMAIL_ONLY"] = settings.ALLOW_RATINGS_CONFIRMED_EMAIL_ONLY
     ret["VIDEO_PLAYER_FEATURED_VIDEO_ON_INDEX_PAGE"] = settings.VIDEO_PLAYER_FEATURED_VIDEO_ON_INDEX_PAGE
     ret["RSS_URL"] = "/rss"
+    ret["LIVESTREAM_BACKEND"] = settings.LIVESTREAM_BACKEND
```
mgogoulos, line 34: you're not adding a default value on settings, so this will make the process break with 
```
AttributeError: 'Settings' object has no attribute 'LIVESTREAM_BACKEND'
```

Better wrap this in a try/except to ensure that it does not break

mgogoulos:
I would like to things here before I can merge the PR: 
1. Understand what the changes are doing, and how they work
2. Make sure that a new (or existing) installation that is NOT using this feature, does not get affected at all.

files/context_processors.py
KyleMaas: Added in latest commit

mgogoulos:
thanks for the updates, will try to test it out, and mostly ensure about 2. (Make sure that a new (or existing) installation that is NOT using this feature, does not get affected at all.), thanks!

KyleMaas:
I'll note that I'm still running this, and it still works well, even with newer versions of Owncast (0.1.2).

SerJaimeLannister:
Any updates on this since its genuinely useful for me!

mgogoulos:
I'm closing this as we are working towards an integration with Owncast with more advanced options. However this is a good start and reference

PR #829, opened 2023-06-29
Disable encoding and show only original file

mgogoulos:
## Description
https://github.com/mediacms-io/mediacms/discussions/285


mgogoulos:
Things to test: if `settings.DO_NOT_TRANSCODE_VIDEO` is enough where it is placed, or whether it needs to be in other places (eg inside save). 
Also add some docs

PR #836, opened 2023-07-07
Keep std vertical resolution and extend horizontal if input is not 16/9

morbificagent:


morbificagent:
Hi mgogoulos,

16/9 content isnt affected/changed in any way.

This code-change only affects content which isnt 16/9 and want to "expand" over this 16/9-limits.
The original code was scaling the video so it fits in a 16/9 "Box"... A 23/9 video would become something like 16/6 or whatever, because it MUST fit in a 16/9 box. So the vertical resolution is changed to strange values and is displayed loke described. With this change it can extend horizontal, keeping the vertical resolution in standard 480, 720, 1080,... formats.

mgogoulos:
Hey didn't get the chance to test this out. Wondering if you have a video available that I can try this out (eg if you have uploaded something on demo.mediacms.io). Thanks!

beatboxa:
I know this is a dated issue; but this is an issue for me as well.

@mgogoulos The above solution did not work for me for a video of resolution 4096 × 1716.  This is a fairly standard DCI resolution, for 2.4:1 cinema-standard resolution.

(I was expecting that mediaCMS would encode to at least UHD-width (3840-wide) resolution).  "2160" is an active encode profile on my system.  However, the highest it selected was "1080" (1920 x 1080).

I'll just comment that in general, video height is not the sole determinant of optimal resolution; and in fact, we rarely use that even colloquially.  For example "4K" usually refers to any resolution where the longer side has a resolution of approximately 4000 pixels.  ie. "2160" in mediaCMS refers to what most people colloquially refer to as "4K."

Please let me know if you want me to start a new feature request:  my recommendation would be either ability to define custom resolutions in the UI (which would provide more options for determining container sizes); or a width-based encoding check rather than height-based.

Also, I'm not following the logic of lines 541 & 542 (and apologies...my programming language skills are not the best).  But it seems to be a conditional statement; however, both conditions appear to be the same.  I would have expected row 542 to reverse the condition.

ie. from:
`f"if(lt(iw\\,ih)\\,{target_width}\\,{target_height})",`
to
`f"if(lt(ih\\,iw)\\,{target_width}\\,{target_height})",`

On row 542, in contrast to row 541.  ?

beatboxa:
FYI, by _additionally_ changing row 542 to the above (in addition to row 543), it appears to have worked in my system.

After looking up a few functions, I believe that logic was stating:
- If the input width is less than the input height, then use the target width.  Otherwise, use the target height.

And I believe the change I made restated the logic such that if the height is less than the width, then use the width.  I believe this would be the correct logic to use.

mgogoulos:
hi @beatboxa , feel free to submit a new PR with the resolution, and ideally upload some video so the logic can be tested in each case!

mgogoulos:
This has become convoluted and is hard to follow. Please resubmit issue or PR and let's keep up the discussion there, thanks a lot to all participants

Issue #852, opened 2023-08-09
Large uploads 'fail' but have succeeded


platima:
**Setup**

- Hyper-V guest with 8 cores, 12GB RAM
- 7TB SAS RAID1 drive array
- Vanilla Ubuntu 22 Server install
- All updates installed
- Vanilla mediaCMS 3 as per instructions
- Bare minimal config changes to make it useable

I've found many videos, usually ones over 1GB, upload successfully but show as failed, eg:
![image](https://github.com/mediacms-io/mediacms/assets/13729856/1a3cf070-10c2-4e8a-aec1-b1733ec471aa)
Retry does nothing on this, unlike one that has failed a portion of the way through.

I have tried reducing the chunk sizing;
```
CHUNKIZE_VIDEO_DURATION = 60 * 2
VIDEO_CHUNKS_DURATION = 60 * 1
```
But that did not help

I can repro this easily, but am unsure which logs to check.

platima:
Narrowed this down to storage speed.

I moved it from 2Gbps SAN to 6Gbps DAS and the issue went away.

I could push/pull ~100MB/s synchronously on this connection, and the uploads are limited to my 40Mbps / 5MB/s throughput, so must be something to do with latency sensitivity, or maybe checksumming?

platima:
Tested more and it's not storage, but I was close. It's the upload speed.

If I upload to it from a 40Mbps-up connection, I get this issue. If I upload from a 100Mbps-up connection, it's fine. (the server itself is on gigabit enterprise fibre)

Most services in Australia max out at 40Mbps upload, and that's only if they have fibre or REALLY good copper, so this would likely affect many people.

platima:
Okay with even bigger files it still fails on the 100Mbps connection; 
![image](https://github.com/mediacms-io/mediacms/assets/13729856/795facb8-2cac-4f9b-a448-90667c8f3070)

'Retry' does nothing. Chrome console logs;
![image](https://github.com/mediacms-io/mediacms/assets/13729856/5910a7a8-c91b-4ac3-9d93-941e96f33a07)

I can copy and paste the file to another location, so read/write is fine, and having tried three times it keeps doing the same thing.

The files on the server are still in a few parts too;
`root@mediacms:/home/mediacms.io/mediacms/media_files/original/user/admin# ls -Alh *C036*
[...]

PR #856, opened 2023-08-17
Make short task time limit configureable

tobocop2:
## Description

The short celery task timeout is hard coded in the daemon and also in the task definition. This should just use the newly introduced setting that can be configured and overwritten with the local_settings.py file 

## Steps

*Pre-deploy*

*Post-deploy*

@mgogoulos


mgogoulos:
Thanks a lot for this PR!

Read the discussion on https://github.com/mediacms-io/mediacms/issues/852 , I think a default value has to be there (on the place that the celery command is run) to act as a fail safe, in case a task runs that takes too much time. Plus you get to override this default setting on any specific task you want (cannot find the doc now but it would be easy to make a test, eg run the celery process with a default of 10seconds, run a `hello world & sleep` task with something else and see which one is stronger. Pretty sure that the overriden one.

mgogoulos:
Not sure how relevant this is now, and apologies for not reviewing earlier. However the `chunkize_media` function is only run on new video uploads, and it can take time, so I wouldn't suggest this getting very low.

Issue #845, opened 2023-07-21
Arrangement of subtitles


cdorsat:
**Describe the issue**
Subtitles should be arranged in the order of the first letter of the language, which is more in line with human habits

![subtitle](https://github.com/mediacms-io/mediacms/assets/22068633/5c6bea3c-09ec-4e6f-adfa-885a81da32f8)

sonukumar7258:
Is this issue available, I want to contribute on this?

mgogoulos:
Sure, feel free to open a PR if you want. A hint here, files.models.Subtitle is not having a default ordering, (eg as Media or others). So Subtitles are returned based on their id (what was created first). files.models.subtitles_info can be changed, to include the ordering, eg something like

change:
```
for subtitle in self.subtitles.all():
```

to:
```
for subtitle in self.subtitles.filter().order_by('language__title'):
```

sonukumar7258:
or we can sort language title by first letter also like as stated/ asked by [cdorsat]
Like implementing the sorted function 

like change

` ` `
for subtitle in self.subtitles.all():
` ` `

to : 

` ` `
 sorted_subtitles = sorted(self.subtitles.all(), key=lambda s: s.language.title[0])
 for subtitle in sorted_subtitles:
` ` ` 

Moreever i have created pull request please review and let me know if i have to make some more changes also. 

Thanks for giving me this opportunity and hint.

Regards, 

Sonu Kumar

PR #863, opened 2023-08-28
#845 Issue : Re-Arranged Subtitles by first letter of language

sonukumar7258:
## Description
Simply Sorted the order of subtitles by the first letter of any language.

## Steps

- Simply Created an other variable named sorted_subtitles
- assigned return of sorted function which will take all the subtitles and condition on basis sorting will be performed
- then replace original subtitles with sorted_subtitles

### DIFF
--- a/files/models.py
+++ b/files/models.py
@@ -775,7 +775,11 @@ class Media(models.Model):
         """
 
         ret = []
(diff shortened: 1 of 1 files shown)


mgogoulos:
that works!

PR #871, opened 2023-09-03
Add support for YugabyteDB as a database backend

KyleMaas:
## Description
This adds support for YugabyteDB as a database backend using the Django compatibility module from [yb-django](https://github.com/yugabyte/yb-django).  This allows for running the database in a far more distributed and durable fashion for minimal downtime.

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
Thanks for opening, I'll have to close this as out of context though, as it is very specific to a platform that is not supported on the main branch

PR #881, opened 2023-09-17
Add CAPTCHA to Contact form

KyleMaas:
## Description
One of my installs receives a lot of spam through the Contact form.  This adds a CAPTCHA to the Contact form to protect against spam.

## Steps

*Pre-deploy*
(EDIT) If upgrading an existing installation, make sure to run `pip install -r requirements.txt` to install the CAPTCHA module.

Per the documentation at:

https://django-simple-captcha.readthedocs.io/en/latest/usage.html#adding-to-a-form

You *have* to run `python manage.py migrate` for this to work.  It adds information to the database that the CAPTCHA plugin requires.  Without it, the form will not work.  But this is standard procedure for MediaCMS version upgrades anyway so I don't see it as a major issue.

*Post-deploy*

### DIFF
--- a/cms/settings.py
+++ b/cms/settings.py
@@ -299,6 +299,7 @@ INSTALLED_APPS = [
     "djcelery_email",
     "ckeditor",
     "drf_yasg",
+    "captcha",
 ]
 
 MIDDLEWARE = [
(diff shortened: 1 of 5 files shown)


mgogoulos:
Thanks for the PR but I will decline it, as it is too specific!

PR #883, opened 2023-09-17
Add admin action to generate missing encodings for a particular Media

### LINKED ISSUE #874: Feature to allow an advanced user (or admin) to re-transcode a video
**Describe the feature you'd like**
I've had several instances where videos did not transcode correctly but left all of their metadata in the database.  This results in videos which show up in the listings but crash with a "video files not found" kind of error from the player when you try to view them.  I've had this happen when a server needs to be rebooted for updates while a transcoding job is happening but also under normal operation with nothing special going on.  When this happens, it also tends to crash the entire media viewing page, so my way of dealing with this has been to go into the MediaCMS Administration and delete the media record and reupload it.  It would be nice if, instead, there was something where I could click a button to trigger a re-transcode of the original media.

**Describe alternatives you've considered**
Well, fixing every bug in existance.  Frankly this would be a workaround for transcoding just working every single time without fail, which I don't see as being realistic.

**Additional context**
No screenshots available, since this would be a new feature.

KyleMaas:
## Description

This adds an action to the MediaCMS Administration section to allow you to select Media items and generate missing encodings.  That way, if for some reason an encoding fails or if you enable other encoding profiles (such as enabling VP9 profiles after a video is posted), you can reprocess the media to fill in the gaps.

Fixes #874

## Steps

*Pre-deploy*

*Post-deploy*

(EDIT) Typo

### DIFF
--- a/files/admin.py
+++ b/files/admin.py
@@ -40,6 +40,12 @@ class MediaAdmin(admin.ModelAdmin):
     def get_comments_count(self, obj):
         return obj.comments.count()
 
+    @admin.action(description="Generate missing encoding(s)", permissions=["change"])
+    def generate_missing_encodings(modeladmin, request, queryset):
+        for m in queryset:
+            m.encode()
+
+    actions = [generate_missing_encodings]
     get_comments_count.short_description = "Comments count"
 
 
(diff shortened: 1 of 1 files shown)


tobocop2:
I am having this issue as well, I think this should be merged in asap.

KyleMaas:
Thanks!

PR #885, opened 2023-09-18
Add useful fields to the Encodings admin screen

KyleMaas:
## Description

When looking at the admin screen for Encodings, there is almost no information presented until you click on a particular encoding.  This adds a few useful fields to that screen as additional columns and filters.

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
Useful!

KyleMaas:
Thanks!

PR #886, opened 2023-09-20
Add configuration to set how many threads each ffmpeg process can use

KyleMaas:
## Description

I was running into issues where, while transcoding, a single server would get a load average of over 4x the number of cores it had.  This turned out to be a combination of Celery running as many tasks concurrently as it could and each ffmpeg process using all cores.  This would result in a ton of chunks' transcoding processes competing for CPU, which would cause them to time out.  This change allows you to at least limit the number of threads that ffmpeg can use concurrently so you can actually tune transcoding CPU usage properly.

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/cms/settings.py
+++ b/cms/settings.py
@@ -160,6 +160,7 @@ HLS_DIR = os.path.join(MEDIA_ROOT, "hls/")
 
 FFMPEG_COMMAND = "ffmpeg"  # this is the path
 FFPROBE_COMMAND = "ffprobe"  # this is the path
+FFMPEG_THREADS = 0  # default - use optimal number of threads
 MP4HLS = "mp4hls"
 
 MASK_IPS_FOR_ACTIONS = True
--- a/files/helpers.py
+++ b/files/helpers.py
@@ -554,6 +554,8 @@ def get_base_ffmpeg_command(
 
     base_cmd = [
         settings.FFMPEG_COMMAND,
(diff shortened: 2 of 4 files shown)


tobocop2:
I noticed this as well, I think this should be merged in as well @mgogoulos

tobocop2:
@KyleMaas i think this threads setting should also be applied to the

`produce_sprite_from_video` 

and this block

```python
    if profile.extension == "gif":
        tf = create_temp_file(suffix=".gif")
        # -ss 5 start from 5 second. -t 25 until 25 sec
        command = [
            settings.FFMPEG_COMMAND,
            "-y",
            "-ss",
            "3",
            "-i",
            media.media_file.path,
            "-hide_banner",
            "-vf",
            "scale=344:-1:flags=lanczos,fps=1",
            "-t",
            "25",
            "-f",
            "gif",
            tf,
        ]
```

in `tasks.py`

KyleMaas:
@tobocop2 Good idea.  Implemented.

dogweather:
Ping

mgogoulos:
I'd suggest we don't change the default behavior, but only make use of the -threads when you do explicitly set it on the settings. So keep the default behavior of not specifying it.

 Feel free to resubmit the PR after fixing conflicts too

PR #909, opened 2023-11-08
Changes to support Debian 12 bookworm and PostgreSQL 15

Lightrider87:
## Description
I've tried to install mediacms to a fresh debian VM and figured out that mediacms do not support PostgreSQL 15. In order to do so I had to update the install.sh and cms/settings.py.

With the Update of PostgreSQL 14 to 15 the public Schema is not writable by default and it is not recommended to make it writable. The following link discribes the changes that are necessary for django to work with PostgreSQL 15.

https://gist.github.com/axelbdt/74898d80ceee51b69a16b575345e8457

## Steps

*Pre-deploy*
Make Sure that PostgreSQL 15 will be installed.

*Post-deploy*

### DIFF
--- a/cms/settings.py
+++ b/cms/settings.py
@@ -387,6 +387,9 @@ LOGGING = {
 DATABASES = {
     "default": {
         "ENGINE": "django.db.backends.postgresql",
+        "OPTIONS": {
+                "options": "-c search_path=mediacms"
+        },
         "NAME": "mediacms",
         "HOST": "127.0.0.1",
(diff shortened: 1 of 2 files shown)


mgogoulos:
Hi, thanks for the PR, while this is working for Debian 12, what will be the behavior for the rest if the supported OS on it, as Ubuntu 20/22?

TheFlipside:
thank you so much, this issue stalled my project for days now and with your information i was able to solve it

Issue #417, opened 2022-02-19
ffprobe outputs an invalid integer in some cases


tschig:
**Describe the issue**
I am trying to upload video.mkv File with ~200mb.
When I Upload it via "https://mediacms.exmaple.com/upload" the fineupload "?done" request fails with Status Code 400 and Response:
```
Request URL: https://mediacms.exmaple.com/fu/upload/?done
Request Method: POST
Status Code: 400
{"success": false, "error": "Error with File Uploading"}
```

On the server, the file video.mkv is present and does have the same md5 as the file on my local harddrive.

Server:
```bash
root@d0dd55e3af29:/home/mediacms.io/mediacms/media_files/uploads/1f2ca7a0-167b-42a8-9a43-84a54e85b406# pwd
/home/mediacms.io/mediacms/media_files/uploads/1f2ca7a0-167b-42a8-9a43-84a54e85b406
root@d0dd55e3af29:/home/mediacms.io/mediacms/media_files/uploads/1f2ca7a0-167b-42a8-9a43-84a54e85b406# ls -al
total 204236
drwxr-xr-x 2 www-data www-data       4096 Feb 19 10:49 .
drwxr-xr-x 9 www-data tempgroup      4096 Feb 19 10:40 ..
-rw-r--r-- 1 www-data www-data  209126313 Feb 19 10:40 video.mkv
-rw-r--r-- 1 www-data www-data          0 Feb 19 10:49 video_8pZlTQB.mkv
-rw-r--r-- 1 www-data www-data          0 Feb 19 10:41 video_aqUqPbb.mkv
-rw-r--r-- 1 www-data www-data          0 Feb 19 10:41 video_oLhah4M.mkv
root@d0dd55e3af29:/home/mediacms.io/mediacms/media_files/uploads/1f2ca7a0-167b-42a8-9a43-84a54e85b406# md5sum video.mkv
0a1dcb25b162786b87826b95be394584  video.mkv
```

Local:
```
C:\Users\justi>md5sum video.mkv
0a1dcb25b162786b87826b95be394584 *video.mkv
```

So upload was correct. 
When trying the same with a smaller video file, everything works.
When trying the same file with a Single Server installation in a VM everything works

Log from Docker Container "web":
```

[...]

PR #925, opened 2023-12-02
replaced pipe with empty string

tdsotm:
## Description
Issue was seen before here:
https://github.com/mediacms-io/mediacms/issues/417

I applied the same fix and it solved my issue, so I thought a PR on the main repo would be helpfull.


mgogoulos:
awesome!

PR #928, opened 2023-12-02
Add a workaround for the start time race condition in Video.js

### LINKED ISSUE #926: Linking to a particular time doesn't work
**Describe the issue**
See this discussion:

https://github.com/mediacms-io/mediacms/discussions/319

I've received user complaints and have tried this myself and confirm that it does not work.  The UI elements for generating a timed link in the Share button work, but when you actually go to the link it doesn't work.

**To Reproduce**
Steps to reproduce the issue:
1. Go to a video
2. Skip partway through the video
3. Click Share
4. Check the "Start at XX:XX" checkbox and copy the link
5. Paste the link into another tab or browser
6. Video plays from the start instead of from the time specified

**Expected behavior**
This:

https://github.com/mediacms-io/mediacms/blob/main/docs/user_docs.md#using-timestamps-for-sharing

 - OS: Has been tested on both Gentoo Linux and Windows from the user side
 - Installation method: single server install
 - Browser, if applicable: Chromium and Firefox, I believe

KyleMaas:
## Description
Turns out the timestamp problem is due to a race condition within Video.js.  Since we can't really fix that without breaking the ability to update Video.js in the future, this adds a workaround for the race condition so that linking to a particular time in a video works.

Fixes #926

See also https://github.com/mediacms-io/mediacms/discussions/319

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/frontend/src/static/js/components/video-player/VideoPlayer.jsx
+++ b/frontend/src/static/js/components/video-player/VideoPlayer.jsx
@@ -197,6 +197,8 @@ export function VideoPlayer(props) {
       const paramT = Number(urlParams.get('t'));
       const timestamp = !isNaN(paramT) ? paramT : 0;
       player.player.currentTime(timestamp);
+      // This is a really hacky way to work around a race condition within Video.js where it thinks it was ready in currentTime() but it actually wasn't so it clears this value:
+      player.player.cache_.initTime = timestamp;
     });
 
     return () => {


Adito5393:
I just tried your fix and is not working as expected... I go to URL`/view?m=5pxaAe2fs&t=15` & the player is still not started at the expected time and when I click the play button, it runs from the begining 

I also re-build all the JS static files and added them to my docker named volume.
Build using Node 14 as recommened in the frontend docs: `npm run dist`
I can see in the browser that the change is available in the final media.js file:
image

Any idea how to get the URL timestamp working?
FYI, I've tested on the [v4.2.0](https://github.com/mediacms-io/mediacms/tree/v4.2.0/)

mgogoulos:
This is now fixed in https://github.com/mediacms-io/mediacms/releases/tag/v4.6.0

PR #938, opened 2023-12-15
Only finalize an encoding when all chunks have finished

### LINKED ISSUE #929: Chunks show encoding a Fail status, but not any apparent problems
**Describe the issue**
Often times when transcoding to an MP4 profile one of the chunks of an encoding will come back as an encoding fail but the video seems to play fine if I pull the actual encoded file.  The logs don't show any errors and look the same as successful chunks, `Retries` is `0`, `Status` is `Fail`, and `Total run time` is also `0`.  But then if I click on `Encoding file` it plays just fine and everything looks okay.  I also never see this with VP9 encoding.

**To Reproduce**
I'm not sure how to reproduce this.  However, I did not see this problem until I moved to a server setup with much higher capacity but also higher disk latency.  I've raised the timeouts for Celery to allow each task to run for considerably longer, far longer than successful chunks take to encode, but the issue persists.

**Expected behavior**
I would expect that the chunk, which from everything I'm seeing did in fact succeed in encoding, would be marked as `Success` instead.  What I find most odd is that `Total run time` is `0` - seems like even for a failure, that should be filled in with some kind of a value.

**Screenshots**
Can't share screenshots of this setup unfortunately.

 - OS: Ubuntu Linux
 - Installation method: single server install
 - Browser, if applicable

**Additional context**

KyleMaas:
## Description
Make sure that all chunks are done before kicking off HLS creation.  Fixes #929

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/files/models.py
+++ b/files/models.py
@@ -1478,7 +1478,7 @@ def encoding_file_save(sender, instance, created, **kwargs):
                     break
 
             for chunk in chunks:
-                if not (chunk.media_file and chunk.media_file.path):
+                if not (chunk.media_file and chunk.media_file.path) or chunk.status != "success":
                     complete = False
                     break
 


KyleMaas:
Hmm...actually, I've only tested this on successes.  It probably doesn't handle cleanup correctly for failures.  Converting this to a draft until I can get that figured out.

KyleMaas:
Okay, I think this should be good now.

KyleMaas:
So, after running this in production for a while, this is still a problem.  This does seem to fix the case where failures happen, but it only runs after the chunk has already been marked as a success, which means if the concatenate and post encode processes take a long time, other chunks may also go into "success" state, see that this one is in "success", and proceed to also concatenate and post encode.  So this still needs some work.  This check really should probably happen on the transition from "running" to "success", not after it has already been marked "success" on the chunk.

mgogoulos:
@KyleMaas thanks for submiting this PR along with the other, I'm struggling to find some time to handle all of them...

as far as this one is related, I'm wondering what are the cases you encounter the problem. Like what types of volumes of video are necessary for it to happen. I haven't seen it myself, but it could be the case that I've missed it, specially if there's not much logging. Perhaps add more log messages to help debug the case?

KyleMaas:
@mgogoulos The main issues come in with long, very large videos.  For example, I've got a server processing one video right now that's about 3 hours long.  Fairly high resolution.  Don't remember the file size, but it's in the several gigabytes range.  That server's been processing that one video for about 22 hours.  And currently it has one Bento operation running and another 30 concurrent instances of `cp -rT` copying to the same output directory.  Usually there are more Bento instances running concurrently as well, but I think this one's just about done processing so now it's mostly just a bunch of copy operations stepping on each other.

KyleMaas:
@mgogoulos Had some more time to track this down further, and I think this does actually fix part of the problem.  But there is also a separate problem I just filed as #962 which is at least as bad if not worse.  So I think this one's still valid and is actually probably correct.

mgogoulos:
I'll need to have another look on this, as a quick fix I've commited https://github.com/mediacms-io/mediacms/commit/5a1e4f25ed41a6f3e6003b3c0afff76acf58d23a and released  v4.2.0

KyleMaas:
@mgogoulos Cool.  Thanks for taking a look at this.  What you did is similar to what I did in this commit:

https://github.com/mediacms-io/mediacms/commit/ea7b4abdc49bda0b6f287d7605badd8cd14b2629

...which I've been running for a while now, and it kinda sorta works but doesn't fully fix the problem.  So this is probably still an issue.

mgogoulos:
> @mgogoulos Cool. Thanks for taking a look at this. What you did is similar to what I did in this commit:
> 
> [ea7b4ab](https://github.com/mediacms-io/mediacms/commit/ea7b4abdc49bda0b6f287d7605badd8cd14b2629)
> 
> ...which I've been running for a while now, and it kinda sorta works but doesn't fully fix the problem. So this is probably still an issue.

for sure, it worths to checkout for a better solution, eg check what has changed and produce the HLS files for this profile and only...

mgogoulos:
closing, as this still is not solving the issue on an efficient way, and we're discussing this on another issue

PR #963, opened 2024-01-22
Calculate md5sum on uploads while in /tmp, deduplicate code

### LINKED ISSUE #937: md5sum on uploaded videos before moving them from temporary directory
**Describe the issue**
I think this is part of the cause of #852.  I noticed the other day when MediaCMS was processing an upload, it was running an `md5sum` on the uploaded video before it responded to the client with a success message.  In tracking this down, it looks like this is being done here:

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/helpers.py#L273

And going up the call stack:

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/models.py#L459

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/models.py#L430

Which I think is coming from here:

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/models.py#L1373

For those of us with large slow primary video data storage but fast `/tmp` on a `tmpfs`, this really should be happening locally on the server processing the upload rather than making the uploading client wait for it to go to the slow disk and then be read back again for the `md5sum`.

**To Reproduce**
The impact of this is not particularly noticeable unless you're processing huge video files and your storage is slow, so this may not be easy to duplicate without uploading videos >1 hour long.  However, if you can set up the conditions correctly:

1. Upload a long video - at least an hour
2. In a separate window, open a terminal and watch `ps` or `top` for an `md5sum` command
3. Notice that this happens very early in the process

**Expected behavior**
The `md5sum` should be calculated on the uploaded file before being moved from `/tmp` to primary storage.

**Screenshots**
N/A

 - OS: Ubuntu Linux
 - Installation method: Single server install
 - Browser, if applicable: N/A

KyleMaas:
## Description
Fixes #937 

Also deduplicates the code for calculating md5sums, since that was being done separately in multiple places in the code

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/files/helpers.py
+++ b/files/helpers.py
@@ -270,13 +270,6 @@ def media_file_info(input_file):
         ret["fail"] = True
         return ret
 
-    cmd = ["md5sum", input_file]
-    stdout = run_command(cmd).get("out")
-    if stdout:
-        md5sum = stdout.split()[0]
-    else:
-        md5sum = ""
-
     cmd = [
         settings.FFPROBE_COMMAND,
         "-loglevel",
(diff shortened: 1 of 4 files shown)


mgogoulos:
with https://github.com/mediacms-io/mediacms/pull/1187 I've made the md5sum calculation redundant by default, since it is problematic in huge files taking time to run, plus the fact that it is "just" stored in the Media model and is not in use. Sorry for not merging this one first, please feel free to re-open after syncing with latest main!

Issue #1002, opened 2024-04-17
 [119] FATAL:  role "root" does not exist


Futureganteng:
Hello, sorry i was not familiar with postgre

i get these error when i run the docker-compose up
db_1             | 2024-04-17 22:51:02.375 BST [119] FATAL:  role "root" does not exist
i logged on as root on the console

then i've logged on as admin on the website but
i get these message and the mediacms website goes down
celery_worker_1  | 2024-04-17 21:51:14,234 INFO reaped unknown pid 54 (terminated by SIGKILL)

Nouri-redouane:
I got the same message, and i solved that by adding the following arguments to the "pg_isready" command in docker-compose.yaml file:
![image](https://github.com/mediacms-io/mediacms/assets/100516721/24ac125e-ada3-4d17-b71a-580dac299b75)
because the pg_isready command without -U is going to use the system user as the database user (in my case "root" because i executed docker-compose using sudo) 
i got that from:
https://stackoverflow.com/questions/70205657/postgres-docker-role-root-does-not-exist

mgogoulos:
Pr was merged and this is fixed now!

mgogoulos:
Pr was merged and this is fixed now!

Issue #894, opened 2023-10-05
db_1 FATAL:  role "root" does not exist


urza:
I installed mediacms via docker compose on ubuntu server 22.04.
I see a lot of these in output:

`db_1             | 2023-10-04 18:53:50.517 BST [61499] FATAL:  role "root" does not exist`

Is this something I need to fix somehow or is it ok to ignore?

mgogoulos:
This is a message that can be ignored, but since it's spamming logs should be removed (fix has to be trivial)

vijaykumar1710:
@mgogoulos is there way we can get rid of this and use the app locally on macbook? serving locally also from docker only.

joon612:
Same issue here, is there any update?

makerduck:
As mentioned in #1002 there's a fix from @Nouri-redouane to suppress the output:
in the docker-compose.yaml got to "db: healthcheck:" and edit `"pg_isready"` to 

`"pg_isready -d $${POSTGRES_DB} -U $${POSTGRES_USER}"`

worked fine for me

working-name:
@makerduck that might work with the defaults but not if you actually bothered to change the env vars. Now getting unknown role for whatever I called the POSTGRES_USER which is `poopoo`:

```
mediacms-db-1             | 2024-06-10 04:08:03.095 BST [107] FATAL:  role "poopoo" does not exist
```

So I think this points to an issue in the setup process as well.

mgogoulos:
Pr was merged and this is fixed now!

PR #1029, opened 2024-06-01
Fix postgres role output

makerduck:
Fixes "garbage" output: 
`db_1 FATAL: role "root" does not exist` 

as discussed in #1002 and #894 
first proposed by @Nouri-redouane 

## Steps
*Pre-deploy*

*Post-deploy*


mgogoulos:
Thanks for the fix!

PR #1036, opened 2024-07-03
Images & dependencies update

laurfb:
## Description

Update some image & python packages; mitigate some CVE. 
Rise max upload file size to 100G. Remove nginx file size limitation.

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
Hey, I'd like to merge this but first want to ask you whether you want to update with even more recent versions, eg latest Django version. If not, I can still merge it and work towards updating these myself. 

Question, why are you resetting the client_max_body_size parameter? 
Also I think the setting that sets a 5GB limit needs to be there, as is it a sane default value. It could be risky to set to unlimited for a generic installation. can you revert it plz?

laurfb:
Hi.

Some (more) recent version update broke some mediacms functionality... I went up to the maximum possible version which did not block/alter mediacms functionality.

Regarding client_max_body_size – sometime I need to upload large clips (10-20-50GB). At least on my case. :)
But I think you're right, maybe it's better to let everyone decide for themselves, with particular settings. Not everyone needs that.

 

I wonder if you have/will have enough time for a new upload app, for large clips and external storage management (clips already stored in external NAS/SMB shares). 😊

 

Thanks you for your hard work on mediacms!

 

Laur

mgogoulos:
Hi, formatting broke so I'll attempt to discuss all issues

> Hi. Some (more) recent version update broke some mediacms functionality... I went up to the maximum possible version which did not block/alter mediacms functionality. 

Do you remember which these libraries/systems have been?

>  large clips and external storage management 
I haven't tried to upload large files but I would assume that even the current uploader should be able to handle them. I'd like to change the uploader with uppy.io since it's modern + allows for passing some metadata as the file path, which can be used to construct a structure on the system, when uploading a folder. 
also external storage is a big topic by itself, there have been attempts with S3 and there are installations that utilize it, it's just that the PRs didn't find their way in main... 

Feel free to share what you are building/working on!

mgogoulos:
I've performed updates to latest versions in https://github.com/mediacms-io/mediacms/pull/1187 , please synd and feel free to re-open

thebigcorporation:
I might be seeing migration issues with postgres.

mgogoulos:
I've added a note on the https://github.com/mediacms-io/mediacms/pull/1187 (i had that note on the release). It's a major version update from PG 15 to PG 17 and it is not compatible. There are two solutions, either edit the local docker-compose and keep PG 15, or you will have to perform a db_dump and db_restore to get it working to PG17. I'd say that in most cases staying in PG15 is perfectly fine. 

Unless if you mean something else here, eg on a new installation, there shouldn't be issues there...

thanks

PR #1052, opened 2024-09-05
fix #943

### LINKED ISSUE #943: Unable to automatically clean up residual files
The encoded files are not cleaned up after the source file is deleted.
and old thumbnails will not be deleted after updating the thumbnails.

thau0x01:
## Description
Fixes issue #943 and other issues related to chown bug while running the containers on macos.

Updated the `deploy/docker/entrypoint.sh` file to ignore the `.git` directory mounted by the volume into the application's root dir. This avoids the permission errors mentioned in several issues. 

This is the new command.
```bash
find /home/mediacms.io/ ! \( -path "*.git*" \) -and ! \( -user www-data -group $TARGET_GID \) -exec chown www-data:$TARGET_GID {} +
```

## Steps

*Pre-deploy*
N/A
*Post-deploy*
N/A

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -28,7 +28,8 @@ else
 fi
 
 # We should do this only for folders that have a different owner, since it is an expensive operation
-find /home/mediacms.io/ ! \( -user www-data -group $TARGET_GID \) -exec chown www-data:$TARGET_GID {} +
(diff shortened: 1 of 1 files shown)


mgogoulos:
Thanks a lot! I will merge this but then will merge another PR that simplifies the find command even further. Would be great if you can test it

PR #1075, opened 2024-10-04
Translations

mgogoulos:
## Description
A long requested feature is here: Translations!!!


KyleMaas:
Wow...that was a huge one.

mgogoulos:
> Wow...that was a huge one.

true...i had started that work some months ago but it wasn't completed. I was able to resume it this week, and brought it to a good enough state, and thought it was better to merge and improve in the future than wait until 100% coverage is performed. 

Most of the changes were inevitably on the frontend directory because apparently the strings were not marked as ready to translate.  There will be other strings that need translation, but there is foundation for this (a. mark them as translatable with translateString, b. add the strings to translate on en.py, c. run the command process_translations that will populate the strings to other languages too, d. translate the strings) .

Also I didn' t translate the edit forms (as edit media) but this is easy to do as well, and these are pages that less people see anyways. 

And as far as the translated strings themselves, i'm going to check check any suggestions or PRs and add that feedback so they are improved.

PR #1086, opened 2024-10-18
Set option for allowing only certain domains to register

yatesdr:
1) Create a key ALLOWED_DOMAINS_FOR_USER_REGISTRATION, default as an empty list in settings.py
2) Apply the key in the E-mail registration validation function to supersede the RESTRICTED_DOMAINS.... setting only if there are items in the new list.


cms/settings.py
```diff
@@ -129,6 +129,7 @@
 USERS_CAN_SELF_REGISTER = True
 
 RESTRICTED_DOMAINS_FOR_USER_REGISTRATION = ["xxx.com", "emaildomainwhatever.com"]
+ALLOWED_DOMAINS_FOR_USER_REGISTRATION = []
```
mgogoulos, line 132: add a comment, eg 
# comma separated list of domains, eg ['domain1.org', 'domain2.org']

users/adapter.py
```diff
@@ -10,6 +10,10 @@ def get_email_confirmation_url_stub(self, request, emailconfirmation):
         return settings.SSL_FRONTEND_HOST + url
 
     def clean_email(self, email):
+        if len(settings.ALLOWED_DOMAINS_FOR_USER_REGISTRATION):
```
mgogoulos, line 13: I'd suggest to query for this more defensively, eg like this
```

if hasattr(settings, "ALLOWED_DOMAINS_FOR_USER_REGISTRATION") ...
```

this ensures that `settings` won't break with an `AttributeError` in case it's missing in settings.py

Also no need to check with len() explicitly, the same can be through:

```
if hasattr(settings, "ALLOWED_DOMAINS_FOR_USER_REGISTRATION") and settings.ALLOWED_DOMAINS_FOR_USER_REGISTRATION: 
```

mgogoulos:
Great, I made some minor suggestions, if you can check them. Also can you add a small entry in https://github.com/mediacms-io/mediacms/blob/main/docs/admins_docs.md ? This doc is not a good way of keeping documentation but it's better than nothing :)

cms/settings.py
yatesdr: Complete.

mgogoulos:
Merged, thanks for the PR, feel free to contribute to other parts too :)

yatesdr:
> Merged, thanks for the PR, feel free to contribute to other parts too :)

Happy to contribute, there's a couple other minor features I'm examining, but not sure if they'd be of use or if I have the skillset to properly implement them.   

My current use-case is as a back-end media manager to Canvas, an online learning platform, to host videos that are often not licensed to be shared publicly.  To facilitate this use-case, the two main features I'm looking into are:

1) Create an option for a "clean embed", with just a play button.   The default embed player includes social aspects I'd prefer not to have in the corners such as share and the video information.    Currently, it's easy enough to override these features in the CSS bundle and it's good enough, but it may be useful for other people to also have this available as a configuration or separate "clean embed" link that simply provides the clean stream and basic controls.

2) Investigate adding some type of auth protection for streams to help limit external sharing of private content.  This is necessary because a number of the videos we use for training are copyrighted, and simply having them unlisted is a potential issue.   It of course works fine if GLOBAL_LOGIN is enabled, but when integrating with a 3rd party application that can't directly log-in I'm not clear on if there's currently a way to permit the stream only to the intended application.   The log-in and auth for end-users of the educational platform is already handled, so I believe setting an endpoint and auth token from the platform may be enough to accomplish this, but don't have a good bead on how to integrate it yet.

Thanks again!

D

twhite96:
I'm struggling to understand where to put these types of environment variables in `deploy/docker/local_settings.py`. Do I need to use `os.getenv` on each variable or can I just place the variables at the bottom of the file? And example:
CleanShot 2025-03-20 at 19 49 37 copy

PR #1092, opened 2024-10-23
Permit embed with global login

yatesdr:
1).   Add a settings key to optionally allow embedding videos from a list of trusted referers when GLOBAL_LOGIN_REQUIRED is true.   This is useful if you wish to use MediaCMS as a back-end content manager, with the embedded files being used on another site.

2).  If above option is set, modify the auth-bypass regex list in settings.py to permit loading from all origins in the embed view without auth, to paths in r'^/embed.\*' and r'^/media/.\*'

3). Modify embed view to validate embed requests against the list of allowed referers - check request.META['HTTP_REFERER'], parse the domain, and compare to the list.   This required including urlparse to perform the domain parse in a reliable and standard way in this view.

4). Create a "fake player" to render (templates/cms/embed-403.html) with a mock 403 error, that fits to parent element and is generally styled like a vis player element.   This is required since embedding in rich content editors often does not properly set referrer on otherwise trusted sites, and giving a 500 API error or redirecting to the MediaCMS main login page is not desirable in this use case.

Example "Bad referrer" render in the stock embed iframe:
image

5).  Doc updates to reflect this new option and its usage.


yatesdr:
Pull request contained other content, will re-issue.

mgogoulos:
ping

yatesdr:
Check in draft PR #1093, it pulled a new # when I opened just these changes as a draft.   Some of the linting is still not passing and I'm not sure how to fix it, but it seems unrelated to my PR.

mgogoulos:
thanks, I'll have a look as soon as I can :)

PR #1093, opened 2024-10-23
Allow embedding by list of trusted referers when GLOBAL_LOGIN is enabled.

yatesdr:
1). Add a settings key to optionally allow embedding videos from a list of trusted referers when GLOBAL_LOGIN_REQUIRED is true. This is useful if you wish to use MediaCMS as a back-end content manager, with the embedded files being used on another site.

2). If above option is set, modify the auth-bypass regex list in settings.py to permit loading from all origins in the embed view without auth, to paths in r'^/embed.\*' and r'^/media/.\*'

3). Modify embed view to validate embed requests against the list of allowed referers - check request.META['HTTP_REFERER'], parse the domain, and compare to the list. This required including urlparse to perform the domain parse in a reliable and standard way in this view.

4). Create a "fake player" to render (templates/cms/embed-403.html) with a mock 403 error, that fits to parent element and is generally styled like a generic player element. This is required since embedding in rich content editors often does not properly set referrer on otherwise trusted sites, and giving a 500 API error or redirecting to the MediaCMS main login page is not desirable in this use case.

Example "Bad referrer" render in the stock embed iframe:
image

5). Doc updates to reflect this new option and its usage.

### DIFF
--- a/cms/settings.py
+++ b/cms/settings.py
@@ -134,6 +134,7 @@ RESTRICTED_DOMAINS_FOR_USER_REGISTRATION = ["xxx.com", "emaildomainwhatever.com"
 # Empty list disables.
 ALLOWED_DOMAINS_FOR_USER_REGISTRATION = []   
 
+
 # django rest settings
 REST_FRAMEWORK = {
     "DEFAULT_AUTHENTICATION_CLASSES": (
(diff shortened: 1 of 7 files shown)


WaiserHai:
any updates ?

yatesdr:
> any updates ?

It’s been working for me in production since the PR was submitted, but I think the PR will not be accepted so just maintain a separate branch now.   Using it with Canvas to embed training videos on the trusted site.

mgogoulos:
I'll have to close this since there has been a refactoring in files/views and it breaks there (plus has other conflicts). Not sure if you're interested to resolve all these and resubmit. Thanks

WaiserHai:
okay so is it now implemented or not ?

yatesdr:
It was closed without adoption per mgogoulos’s comment, and then a refactor was done which altered some of the pages and afaik this is not on the roadmap for the head version.   I am not actively working on it.

If you need this feature I am running the forked branch with it for over a year now, but not maintaining it to current level.

Issue #962, opened 2024-01-22
create_hls being called multiple times on the same video as additional MP4 encodings finish


KyleMaas:
**Describe the issue**
Related to #929 and pull request #938.  Part of the problem appears to be coming from how `create_hls` is written and is called, in that it appears to be running on the same video ID every time *any* MP4 encoding finishes.  Which means if you have 4-5 different resolutions, `create_hls` appears to be intentionally running 4-5 times.  When you have huge videos, this results in atrocious amounts of disk I/O.  We need to figure out a way to get `create_hls` to only run after the very last MP4 encoding has finished.

See here for reference, walking up the call stack to where this is coming from:

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/tasks.py#L408

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/models.py#L642

https://github.com/mediacms-io/mediacms/blob/c5047d8df8686d75100e5099489be4fd1bf5f733/files/models.py#L1564

If you note in that last one, that's ideally not being done per chunk (which is what #938 should fix) but is still done on every successful encoding, which means every enabled MP4 encoding profile should ultimately result in a call to `create_hls` if the encoding is successful.

**To Reproduce**
Steps to reproduce the issue:
1. Make sure your Celery is configured to allow for more concurrent tasks than you have MP4 encoding profiles so that this can all happen concurrently.
2. Upload a video that takes at least an hour to transcode, like a high-resolution video which is several hours long.
3. Watch your tasks in `ps` or `top`.  You should see Bento running multiple times, and if your processing is long enough or your disk access slow enough, you should eventually see those Bento processes finish and the `cp` start building up and colliding.

**Expected behavior**
`create_hls` should only run once per Media record, when all the MP4 encoding processes for that video have finished.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS: Ubuntu Linux
 - Installation method: single server install
 - Browser, if applicable: N/A

**Additional context**
N/A

mgogoulos:
The description is valid, this was made on purpose neglecting the fact that it adds extra overhead on large videos (I hadn't noticed that it takes time/resources, thought it would be a very light process).

KyleMaas:
[...]

Issue #957, opened 2024-01-15
Player wont show all encoding formats


jmoschetti45:
I've setup a few other formats (and removed a few) from the default encoders. The ones I've removed from the encoding list disappear from the player quality options, but the ones I've added don't appear.

The transcoded files exist and play fine manually, and show up in the download list.

All h264 like the defaults, just different resolutions. 1440 and 2160 since I have a lot of 4k and 8k content.

Current git pull from 1/13/2024, everything else works.

mgogoulos:
This is now fixed in v4.2.0, there was indeed an error that is now fixed

PR #1095, opened 2024-10-28
ensure create_hls runs only after all encodings finish

bobcarroll:
## Description
This patch addresses #929 and #962 by ensuring that `create_hls` only runs Bento after all h264 encodings are finished. #957 may be fixed by this too as higher resolution encodings won't appear in the player selection without HLS.

Calling Bento with more than one media file doesn't work because the `files` argument is interpreted as a single (quoted) value. It seems like the original intention may have been to generate HLS files as each encoding finishes, and this case wasn't triggered until I eliminated duplicate runs of Bento.

#938 didn't fix the problem for me. `create_hls` was (indirectly) being called from line 1588 of `files/models.py` and all of my uploads were failing to encode regardless of size. The encoded files were still created and visible in the web UI, with the exception of higher resolution encodings.

## Steps
Upload a video file with Bento installed and properly configured.

### DIFF
--- a/files/tasks.py
+++ b/files/tasks.py
@@ -427,15 +427,24 @@ def create_hls(friendly_token):
 
     p = media.uid.hex
     output_dir = os.path.join(settings.HLS_DIR, p)
-    encodings = media.encodings.filter(profile__extension="mp4", status="success", chunk=False, profile__codec="h264")
+    encodings = media.encodings.filter(profile__extension="mp4", chunk=False, profile__codec="h264")
+
+    if not all((x.status == "success" for x in encodings)):
+        logger.info('not all encodings are finished')
+        return False
+
     if encodings:
         existing_output_dir = None
         if os.path.exists(output_dir):
             existing_output_dir = output_dir
             output_dir = os.path.join(settings.HLS_DIR, p + produce_friendly_token())
-        files = " ".join([f.media_file.path for f in encodings if f.media_file])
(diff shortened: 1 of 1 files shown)


mgogoulos:
I'll need to have another look on this, as a quick fix I've commited  https://github.com/mediacms-io/mediacms/commit/5a1e4f25ed41a6f3e6003b3c0afff76acf58d23a and released  v4.2.0

Can you sync this branch, since there's a conflict to tasks.py? 

Thanks

bobcarroll:
@mgogoulos done

mgogoulos:
this is good work, however i'm thinking what could be the implications, fore example

1. hls will be prepared once all encodings have finished, which means that it will take longer than it currently does, because now it prepares a 240px resolution that is very fast, and makes the video available. If it has to wait fro all resolutions to finish, it could take much more time
2. the fact that it runs every time there's a new encoding read (current behavior), means it acts as a fallback if some encoding task fails. If your changes get merged, we're losing this mechanism and if an encoding goes wrong, there's no HLS

I think what is needed here is a special handling of tasks.create_hls in order to do two things:
1. call Bento with the new files only (not the existing ones), so basically call Bento each time there's a new encoding with this encoding file only
2. update the master.m3u8 file with any new entries...

what we do now, run Bento with all files, can be optimizing with running with specific files only. I think that's a better solution in the long term

mgogoulos:
please sync ith main and re-open if you want!

bobcarroll:
@mgogoulos Sorry I'm just now seeing your note. I haven't had a chance to look into this again. I'll re-open when I can get back to it.

PR #1120, opened 2024-11-18
feat: fullscreen images slideshow

aleensd:
## Steps

*Pre-deploy*

*Post-deploy*


aleensd:
@mgogoulos i still need to clean up the code a bit test it more and add keyboard keys functionalities

mgogoulos:
This looks great so far!

I've made these two changes and pushed them in main, can you sync in order to get them? 

1. the slideshow_items now returns the first item as the image itself. Thus when you click on the image, you get that image as the first fullscreen item. Before this, you would get another image as first, and that would be strange!
2. I've made the number of slideshow items as configurable, checkout `files/models.py` on `slideshow_items = getattr(settings, "SLIDESHOW_ITEMS", 30)`

It expects a setting SLIDESHOW_ITEMS and will read that number, otherwise it sets a default of 30. 
If you want, give it a test, eg upload a larger number of images, and increase this number to see how they look

Other than that, I have the following suggestions
- [ ] on the media page, when you hover on the image, show a message (eg 'load fullscreen'), the indication is good but i think this is also needed
 
- [ ] show the title of media, eg below the image, since we have it on the slideshow_items dictionary

- [ ] I noticed the following: if i visit a page that has a large image, it starts showing, before it completes if i select another one, it will wait until the first is downloaded and show it before it proceeds. Could we somehow cancel the first request, when there is a change? I noticed that requests are canceled when i click 2-3 times right (not sure if you are doing this, or if it's the browser doing it) but this is nice behavior I think and we should include it in the scenario i describe (hope it makes sense btw!). The idea is if we can make the transition a bit smoother, let's also add the title and see how it looks, because you will see there is the title that has changed

mgogoulos:
first level support for image files!

PR #1182, opened 2025-02-02
Do not pin the micro-version in Dockerfile

thebigcorporation:
- the mediacms Docker build needs to pick up security patches
 - this patch removes pinning the base system to 3.11.4, which otherwise probits security patches more recent than 3.11.4 from being picked up by a fresh container build
 - as of the time of this writing, the python-3.11 base has advanced to 3.11.11: see https://github.com/docker-library/python/blob/ded42cf/3.11/bullseye/Dockerfile
 - not declaring the micro version ensures the latest up-to-date base platform image is automatically used
 - in terms of risk we must consider that users are putting this platform on the DMX Internet. The security benefits clearly outweigh any possible (unlikely) minor regressions, since micro version updates are defined as not breaking API/ABI and since this semver based best-practice is closely adhered to in the upstream Linux developer communities.

## Steps

*Pre-deploy*

*Post-deploy*

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.11.4-bookworm AS compile-image
+FROM python:3.11-bookworm AS compile-image
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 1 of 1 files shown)


mgogoulos:
thanks, I used you advice in https://github.com/mediacms-io/mediacms/pull/1187

PR #1187, opened 2025-02-10
Feat lib updates and more

mgogoulos:
This PR performs the following changes:

- update python in Dockerfile
- updates all python libraries (including Django)
- makes md5sum calculation redundant by default
- updates Postgresql used in Docker. For new installations this is ok. For existing ones this is backwards incompatible, and restore through pg_dump+pg_restore has to be performed in order to utilize Postgres 17 (since its not compatible with previous versions and will complain)
- fixes issues with HLS, and adds test to ensure they won't happen again
- allows . and @ in usernames


mgogoulos:
This updates Postgresql used in Docker. For new installations this is ok. For existing ones this is **BACKWARDS INCOMPATIBLE**, and restore through pg_dump+pg_restore has to be performed in order to utilize Postgres 17 (since its not compatible with previous versions and will complain).


The error you are getting is this:
```
FATAL:  database files are incompatible with server
DETAIL:  The data directory was initialized by PostgreSQL version 15, which is not compatible with this version 17.2.
```
Quick solution: edit docker-compose in use and keep previous version (eg 15.2). Otherwise have to perform a dump+restore

update python in Dockerfile
updates all python libraries (including Django)

PR #1196, opened 2025-02-13
Nit: Fix Docker WARN: FromAsCasing

thebigcorporation:
Fixes: L27 'as' and 'FROM' keywords' casing do not match

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
good catch!

PR #1226, opened 2025-03-15
More robust logging and outputs for all exceptions

mrjoshuap:
This is a change to ensure that logging is configured in a sane fashion.  I reworked the configuration so that it will honor LOGLEVEL except when DEBUG is enabled.  Arguments maybe on the best implementation, but I'm running in Kubernetes and want to see warnings+ only.

## Description
Found all exception catches, they now print warnings.  Could argue some might be better as debug/info.  Found missing `import logging` and added appropriate `logger =` setup.

## Steps
With `DEBUG=False`
1. Run MediaCMS, set `DEBUG=False` and `LOGLEVEL` to `[DEBUG|INFO|WARN|ERROR]`
2. Outputs ONLY to console, no files, `LOGLEVEL` and up.
2. Logs should properly be logged, filtered and things will be more obvious.

1. Run MediaCMS, set `DEBUG=True` and `LOGLEVEL` is ignored
2. Outputs to console AND error_files
3. Logs should be properly logged, filtered and things will be way more verbose

### DIFF
--- a/cms/settings.py
+++ b/cms/settings.py
@@ -1,7 +1,10 @@
 import os
+import logging
 
 from celery.schedules import crontab
 from django.utils.translation import gettext_lazy as _
+
+logger = logging.getLogger(__name__)
 
 DEBUG = False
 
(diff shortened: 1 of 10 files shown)


mgogoulos:
Thanks a lot for this PR. There was a big refactoring in views and models in files/, so now everything lives in files/views/ or files/models/ , while before there has been single files (files/views.py and files/models.py). This is why this is breaking with conflicts. 

Can you sync with main, solve conflicts and re-submit? I think it's better if you start from main, and implement the changes again. 

Thanks and regards

PR #1257, opened 2025-04-21
Add Open ID Connect (OIDC) Login

chsasank:
Added OpenID connect login system. I chose OIDC as the provider because it's generic enough and supported by most SSO providers like keycloak, auth0, google etc.

I could get this working only with v4.6 because 5.0 introduced SAML, RBAC etc and complicated things.


chsasank:
Now doing production build and testing.

chsasank:
Production build is also done and I tested it. Here's how sign in looks like now

image

chsasank:
I am happy to modify this PR to make it inline with how project is evolving. Actually getting the template working is what took majority of effort here. settings was quite straight forward.

PR #1274, opened 2025-05-12
One-banger easy install for Ubuntu

yatesdr:
## Description
Provide a single-command installation script for fresh Ubuntu Server installations,  intended to set up a single server with minimum effort.

 - easy-install.sh: The script, adapted from install.sh
 - Doc updates to show usage.
 - requirements.txt - fixed to unpin xmlsec / lxml across Ubuntu 20, 22, and 24 (Python 3.10-3.13 tested)
 - Dockerfile - fixed to incorporate xmlsec /lxml fix and permit image re-build without further headaches.
 - Doc updates to install a modern version of Docker and compose, which appears to be required at this point.

## Notes
The path for the one-liner script as it will eventually sit on github is guessed at the moment, but can be confirmed and tweaked this PR is adopted.


yatesdr:
@mgogoulos - Not sure if you want it or not, but a one-command install can be pretty useful.

yatesdr:
Updated & tested working with Ubuntu 20, 22, and 24.   Changes needed for Ubuntu 20 to work with Django 5.1 were added, including upgrading system python to 3.10 and adding / installing Postgresql 13.

yatesdr:
Eliminated separate requirements-ubuntu-24.txt file.   It turns out it wasn't needed after dealing with the libxmlsec1 problems, so removed. 

For future reference, the conflicts come into play with ubuntu stock libxmlsec1 libraries not matching the expected build of Python 3.10+ builds.   For whatever reason, Ubuntu is using an older version and it needs to be compiled against the local library to properly build.

For 24.04:
```
pip install --no-binary lxml lxml==5.4.0
```

For 20 and 22:
```
pip install --no-binary lxml lxml==4.9.2
```

yatesdr:
Carried fixes for xmlsec and lxml upstream into Dockerfile, so checks will run and pass with new requirements.txt file.

This one should be good to go now!

yatesdr:
- Updated to work with Debian bullseye, bookworm, and trixie
- Tested on Bullseye and Bookworm.

mgogoulos:
Great work, thanks a lot for the PR and for helping other users too. I've prepared two servers and plan to run it for single server and Docker installations. I'm slightly worried that the chances include the Dockerfile but I guess this is unavoidable? 

Will write more after I try it, thanks!

yatesdr:
> Great work, thanks a lot for the PR and for helping other users too. I've prepared two servers and plan to run it for single server and Docker installations. I'm slightly worried that the chances include the Dockerfile but I guess this is unavoidable?
> 
> Will write more after I try it, thanks!

My pleasure, glad I can help!

I tried to keep the Dockerfile changes to a minimum, but unfortunately they were necessary to allow unpinning lxml in requirements.txt - the changes add pkg-config and force compile lxml and xmlsec, and without them build fails.

if you prefer to not do this in the dockerfile, it would require a separate requirements.txt file for docker vs local install, so the minor Dockerfile change seemed to be the lesser evil to me.

yatesdr:
@mgogoulos - I had mistakenly left an install for requirements-ubuntu-24.txt in the submitted patch - fixed now, it was an artifact from earlier testing and it's possible to use a single requirements.txt file.   After moving libxml and xmlsec builds out it is not needed and installs fine for me on all test systems.

docs/admins_docs.md
```diff
@@ -31,22 +31,94 @@
...
 
-## 2. Server Installation
+## 2. Single-Server Installation
 
-The core dependencies are Python3, Django3, Celery, PostgreSQL, Redis, ffmpeg. Any system that can have these dependencies installed, can run MediaCMS. But we strongly suggest installing on Linux Ubuntu (tested on versions 20, 22).
+The core dependencies are Python3, Django, Celery, PostgreSQL, Redis, and ffmpeg. Any system that can have these dependencies installed, can run MediaCMS. But it's strongly suggested to Linux Ubuntu (tested on versions 20, 22, and 24).
 
-Installation on an Ubuntu system with git utility installed should be completed in a few minutes with the following steps.
-Make sure you run it as user root, on a clear system, since the automatic script will install and configure the following services: Celery/PostgreSQL/Redis/Nginx and will override any existing settings.
+## Easy Install
```
mgogoulos, line 38: - make this a subhheader (###)
- even better, add the whole single server installation section with a <details> and <summary>. Its great if one wants to see it but besides this its just too big...

docs/admins_docs.md
```diff
@@ -89,15 +160,32 @@ Database can be backed up with pg_dump and media_files on /home/mediacms.io/medi
 ## Installation
 Install a recent version of [Docker](https://docs.docker.com/get-docker/), and [Docker Compose](https://docs.docker.com/compose/install/).
 
-For Ubuntu 20/22 systems this is:
+For Ubuntu 20/22/24 systems you can add the docker repo:
```
mgogoulos, line 164: i suggest that you bring back the one liner, but remove the docker compose part, since it's now part of docker
Then maybe use `<details> and <summary>` here as well for the manual instructions? so that the actual parts are very clear and lead fast to running them

docs/admins_docs.md
yatesdr: Updates committed to condense to a reasonable size, while still retaining core information.   Please check and advise.

yatesdr:
Ubuntu 22 is still pinned to lxml==4.9.0 which is an older version (5.4 is current).   I'm not exactly clear on why this is not able to upgrade, but will keep working on it and if a fix is possible will submit in a separate PR.   

I think this has to do with Ubuntu 22 using Python 3.10 which uses an older version of xmlsec1-dev, and the breaking changes to lxml > 4.9 in pypy causing the issue.

yatesdr:
> I gave it a try and have to say that it works for Docker, but the single server installation does not start.
> 

Hmm... I tested it on all variants repeatedly before submitting - can you confirm that you were using the updated requirements.txt?   If not - you must also used the patched requirements.txt - the only change is to unpin lxml==5.0.0 and set it to lxml>=4.9.0, otherwise it will reinstall and you get the library version mis-match.

You can either manually tweak requirements.txt or just try the version in my 'easy-install-live' fork, which has committed these changes already to requirements.txt, but runs off the forked code.

yatesdr:
@mgogoulos - The Hezner install went fine for me on Ubuntu 24 using the 'easy-install-live' branch, so I suspect the issue is the requirements.txt update has to coincide with the patch for easy-install.sh

image

image

yatesdr:
@mgogoulos - Patched in a small sed fix to force fix lxml dep for requirements.txt after clone using current requirements.txt.

Runs in Hetzner fine for me with just the one-banger.

yatesdr:
Now tested on Hetzner/Ubuntu 22 and Ubuntu 24 instances, as well as proxmox LXC's of the same.   All builds successful and functional for login, video upload, re-encode and playback.

mgogoulos:
thanks for working on this @yatesdr , I'll try to review and comment or otherwise merge during the week!

mgogoulos:
@yatesdr I'm closing this in favor of https://github.com/mediacms-io/mediacms/pull/1296 that was just merged. I've made as less changes there as possible to ensure it won't be risky for existing installations. 

Please feel free to re-open with any changes you want regarding the documentation, ideally with isolated changes (eg in one file) so they can be merged quicker. Thanks a lot for your work on this and the comments on other issues, please don't feel discouraged for not merging this one.

PR #1279, opened 2025-05-19
Add attribute autocomplete=off to user and password fields

phfuh:
## Description
Add attribute autocomplete=off to user and password fields to improve security in shared computing environments.

## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
hi, have you tested this locally? I'm reading that modern browsers sometimes ignore autocomplete="off" for password fields, and there are suggestions for autocomplete="new-password" / autocomplete="current-password"...

PR #1297, opened 2025-06-16
Fixing date picker in edit media

andyfoundi:
By default, Django uses type=text for the date input, which respects the locale settings. However, when changing the input type to date, it should only accept YYYY-MM-DD format so the input field can be properly handled.

Before:
截圖 2025-06-16 上午11 11 48

After:
截圖 2025-06-16 上午11 12 11


mgogoulos:
thanks for submitting

Issue #1339, opened 2025-08-01
Media titles 100 characters in length are truncated to 99 characters


riyfu:
**Describe the issue**
Setting media titles exactly 100 characters long cuts off the last character.

**To Reproduce**
Steps to reproduce the issue:
1. Go to the View page for any media in a MediaCMS deployment.
2. Click the "EDIT MEDIA" button.
3. Change the title to "Here we have a test video for MediaCMS with a title that is exactly one hundred characters in length".
4. Click "Update Media".
5. Observe the new title on the View page is missing the last character: "Here we have a test video for MediaCMS with a title that is exactly one hundred characters in lengt"

**Expected behavior**
The title should correctly contain all 100 characters.

**Environment (please complete the following information):**
 - OS: Ubuntu Linux 24.04
 - Installation method: Both Docker install and single server install of MediaCMS v6.3.0

**Additional context**

My guess is the problem lies with the `self.title = self.title[:99]` lines in https://github.com/mediacms-io/mediacms/blob/v6.3.0/files/models.py.  These should probably be `self.title = self.title[:100]`.

MeetD99:
Can you assign this to me

mgogoulos:
sure!

MeetD99:
@mgogoulos I have made a PR, please review it

mgogoulos:
thanks for submitting the PR!

PR #1364, opened 2025-09-01
Updated the title splicing length to 100 in playlist, category and media models

MeetD99:
## Description
The title was being spliced as title[:99] which fixed it's length to 99 characters, so I updated it to [:100] in the playlists, categories and media models.

#1339


mgogoulos:
looks good, thanks!

Issue #94, opened 2021-03-11
Respect LOGS_DIR value


timfam:
Mediacms does not respect new value of LOGS_DIR I change location in local_settings.py to = "/var/log/mediacms" yet logs are printed in old logs subfolder

PR #1411, opened 2025-10-18
Fix: Make log directory configurable via environment variable

ArnavB11:
### Description

# [Feature]Respect LOGS_DIR value #94

Small Note: This is my first issue and PR and I found it pretty challenging so if there are any issues with my documentation or in my solution, please do provide your input so that I can do better in the future.

This PR resolves an issue where the `LOGS_DIR` setting, when defined in `local_settings.py`, was being ignored by the Docker environment. The root cause was an order-of-operations problem in `cms/settings.py` where the `LOGGING` dictionary was configured before local settings were imported, preventing any overrides.

This fix ensures that `local_settings.py` is imported *before* the logging configuration is built.

Additionally, this PR introduces several essential fixes to the Docker development environment, particularly for users on Windows with WSL2, to resolve common startup failures. These changes make the local setup process significantly more reliable for new contributors.

**Key Changes:**

1.  **`cms/settings.py`:** Moved the `local_settings.py` import block to a position above the `LOGGING` dictionary. This is the primary fix for the bug.
2.  **`cms/local_settings.py`:** Added logic to read `LOGS_DIR` from an environment variable, aligning with Docker best practices.
3.  **`docker-compose.yaml`:**
    * Configured all application services (`web`, `migrations`, `celery_worker`, `celery_beat`) to build from local source (`build: .`) instead of using a pre-built image. This ensures local code changes are always used.
    * Added the `LOGS_DIR` environment variable to all application services, making the log path configurable.
    * Added a `healthcheck` to the `db` service and `depends_on` conditions to other services to prevent a startup race condition.
4.  **`deploy/docker/entrypoint.sh`:**
    * Added a command to explicitly create the `LOGS_DIR` directory (`mkdir -p`) on container startup. This resolves issues where Docker volume mounts would hide or overwrite the directory.
4.  **`CRLF vs LF Line Endings in Shell Scripts:`:**
    * In your code editor (e.g., VS Code), ensure all shell scripts in the `deploy/docker/` directory (`entrypoint.sh`, `prestart.sh`, etc.) are saved with **LF** line endings before building the Docker image. 

These changes collectively fix the reported issue and dramatically improve the developer setup experience.

---

#### Pre-deploy

No pre-deployment actions are required. These changes are self-contained within the application's configuration and startup scripts.

#### Post-deploy

To verify the fix and demonstrate the new functionality, follow these steps:

1.  **Configure `docker-compose.yaml`:**
    Ensure the `LOGS_DIR` environment variable is set for the relevant services (e.g., `web`, `migrations`).
 
    In docker-compose.yaml, under the 'web' service environment:
    environment:
     
    LOGS_DIR: /home/mediacms.io/mediacms/custom_logs
  

2.  **Perform a Clean Rebuild and Relaunch:**
    Run the following commands in the project root to ensure all changes are applied:
    
    docker compose down --volumes
    docker compose build --no-cache
    docker compose up
    

3.  **Verify Directory Creation:**
    Once the application is running, exec into the `web` container and list the directory contents:
    
    docker compose exec web /bin/bash
    ls -l
    
    The output will now show the `custom_logs` directory, confirming that the setting was correctly applied and the directory was created.

   
image


ArnavB11:
Please do provide your input.

PR #1413, opened 2025-10-20
Video.js fixes and improvements after major upgrade

Yiannistaos:


mgogoulos:
did you see this error happening somewhere / can it be reproduced?

Yiannistaos:
> did you see this error happening somewhere / can it be reproduced?

Which error?

Yiannistaos:
this fixes https://github.com/mediacms-io/mediacms-deic/issues/130

mgogoulos:
i'll test it locally / try to reproduce first, just to make sure it doesnt break other cases

PR #1423, opened 2025-11-01
Add attachment management functionality

christophe195:
## Description
This PR adds the ability to add attachments to a video. 

After uploading, you will find an “attachment” button in the navigation bar when managing the video. Here you can add attachments, name them, and delete them.

Once you have done this, a button will appear next to Download for the attachments.

**Please be sure to check the code in terms of security; I am not very familiar with the project.** 
I did use AI for the translations

image
image


christophe195:
@mgogoulos, Do i need to add you as reviewer after i'm done? If yes, how?

mgogoulos:
Hi @christophe195 , thanks for submitting this PR. 

This looks good. One thing I'd like to mention is that this is a very custom feature, so we don't want to have it there by default for all use cases. Usually we make a configuration for this type of features, that is disabled by default, but you can easily enable it through the setting. Eg that could be a setting ATTACHMENTS_FUNCTIONALITY (or something better!) that is False by default and this check takes place in all areas where a user can see this (like the media page, or the edit media tab) but also in the code. Hope this makes sense!

christophe195:
Sure, will do.

christophe195:
@mgogoulos , option is added

mrjoshuap:
I really like this as a concept, however I have some security related concerns after a casual review on my mobile.. might be off, so let me know if I got something wrong.

- I didn't see a default set of allowed files/mimes to upload or ones to prohibit.
- Should media attachments be subject to the same size restrictions? I say keeping them to under 1MB is way generous.
- What kind of RBAC would be necessary?  Logged in?  Member of group?

PR #1503, opened 2026-04-03
Feat/external video embed

AztecaTech:
## Steps

*Pre-deploy*

*Post-deploy*


mgogoulos:
Closing this as it looks to be 100% AI generated - https://github.com/mediacms-io/mediacms/discussions/1519

If you want to consider this to be merged, I suggest the following:

1. make it smaller, so that it is possible to review. 
2. Do not include files as docs/superpowers/plans/2026-04-02-external-video-embed.md . What's the reason in including these? 
3. this is a feature that might or not interest most of the installations, so it should not affect them anyways. this functionality should be set via a setting, and should be False by default. Then one could enable via setting this to True. Ideally it should live in different files, so that the existing are not affected too much. I can't tell whether it's best to have it on Media model or on a new model, but the idea is that we shouldn't add too many fields there

If you want to get this feature please add some screenshots how you already use it (since I assume you use it) and also let's discuss in the above options, to set some specs.

PR #1530, opened 2026-05-29
feat: support external HLS sources with playlist management and player improvements

alexamirante:
## Summary

This PR adds support for linking external HLS master playlist URLs as media
sources, without requiring a file upload. It also includes several improvements
to the media API, the upload/edit UX, and the video player.

### External HLS support
- New `external_hls_url` URLField on the `Media` model (migration included)
- Media created with an external HLS URL skips local file processing
  (`media_init` is not called)
- The player reads the external URL through `hls_info` and `original_media_url`
- Trim and Chapters controls are disabled in the UI for external HLS media
- Managers can set `duration` for external HLS media via the API
- New upload option in the user dashboard to submit an external HLS URL

### Playlist management via API
- `playlist_ids` field added to media `POST`/`PUT` endpoints, allowing callers
  to assign a media item to one or more playlists in a single request
- The edit-media form now exposes playlist selection and syncs membership on save

### Media API improvements
- `POST`/`PUT` endpoints now accept `enable_comments`, `allow_download`, and
  `tags` fields
- Caption/subtitle upload supported in media update (`PUT`)
- `uploaded_poster` documented in Swagger for `POST`
- Duplicate Swagger decorator and `JSONParser` issues fixed
- Login schema generation conflict resolved

### Player fixes
- Absolute external media URLs handled correctly
- Quality forcing supported for external HLS sources
- Captions control shown for external HLS subtitle tracks

### Regressions fixed
- Subtitle files now pass the protected-media path auth check

### Other
- `docker-compose.yaml`: Admin credentials made configurable via environment variable
  (`ADMIN_USER`, `ADMIN_EMAIL`, `ADMIN_PASSWORD`)
- Python `__pycache__` directories added to `.gitignore`

## Database migrations

A new migration `0019_media_external_hls_url` adds the `external_hls_url`
field to the `files_media` table. It depends on `0018_embedmediacourse` and
fits cleanly into the linear migration graph.

PR #1534, opened 2026-06-03
feat: openid connect

GaijinKa:
### Summary

This PR introduces full OpenID Connect (OIDC) authentication support for MediaCMS via `django-allauth`'s built-in `openid_connect` provider. No new Python packages are required — allauth 65.4.1 already ships the OIDC provider.

The implementation mirrors the existing SAML architecture (`saml_auth/`) and integrates with the `identity_providers` app for role and group mapping.

---

### Motivation

MediaCMS already supports SAML SSO. Several institutional deployments (e.g. IETF, academic networks) require OIDC in addition to or instead of SAML. The goal was to add OIDC support:

- without adding new package dependencies
- keeping the same adapter/model pattern used by `saml_auth/`
- supporting mixed SAML+OIDC deployments on the same instance
- making provider configuration fully declarative and idempotent via a management command

---

### New folder: `oidc_auth/`

**Models** (`oidc_auth/models.py`):

- `OIDCConfiguration` — per-provider configuration linked to an allauth `SocialApp`. Stores claim-to-field mappings (`uid_claim`, `email_claim`, `name_claim`, etc.), behavioural flags (`verified_email`, `email_authentication`, `remove_from_groups`, `save_oidc_response_logs`), and a `scopes` JSONField. On save, `scopes` is automatically synced to `SocialApp.settings["scope"]` so allauth always requests the correct OAuth2 scopes. Enforces one configuration per `SocialApp` via `clean()`.
- `OIDCScopeConfig` — maps a claim to a parser function (stored as a dotted Python path). Unique together on `(oidc_configuration, claim)`. The `ordering` field has been removed — one handler per claim is sufficient.

**Adapter** (`oidc_auth/adapter.py`):

- `OIDCAccountAdapter` — sanitises the username from the `sub` claim (replacing characters disallowed by MediaCMS validators with `_`), then delegates attribute sync and role/group mapping to `perform_user_actions()`.
- `perform_user_actions()` — loads all active `OIDCScopeConfig` rows for the provider, dynamically imports each `parser` function, and calls it with the raw claim value and `parser_options`.
- A `social_account_updated` signal receiver re-runs the same logic on every login, not just the first (keeps roles and groups in sync).

**Parsers** (`oidc_auth/parsers/`):

Two built-in parsers are provided; additional parsers can be added as plain Python functions anywhere on the Python path.

- `profile_picture.sync_profile_picture` — downloads the URL from the `picture` claim and saves it as the user's avatar.
- `ietf_role_mapping.handle_ietf_role_mapping` — maps `[role, group]` claim pairs (joined with `:`) to a global RBAC role and `RBACGroup` memberships, driven by a `role_matcher` dict in `parser_options`. The first matching entry wins.

**Management command** (`sync_oidc_providers`):

Idempotent command that reads `OIDC_PROVIDERS` from settings and creates or updates `SocialApp`, `OIDCConfiguration`, `OIDCScopeConfig`, and `LoginOption` records. Safe to run on every deploy. Supports `--dry-run`.

Example settings block:

```python
OIDC_PROVIDERS = [
    {
        "provider_id": "myprovider",
        "name": "My Provider",
        "client_id": os.getenv("OIDC_CLIENT_ID"),
        "secret": os.getenv("OIDC_CLIENT_SECRET"),
        "server_url": os.getenv("OIDC_SERVER_URL"),
        "oidc_config": {
            "verified_email": True,
        },
        "scopes": ["openid", "email", "profile"],
        "claim_handlers": [
            {
                "claim": "picture",
                "parser": "oidc_auth.parsers.profile_picture.sync_profile_picture",
            },
        ],
    },
]
```

---

### Changes to existing apps

**`identity_providers/admin.py`** — registers the new OIDC models and significantly improves the `SocialApp` admin UX for OIDC providers:
- `get_inlines()`: OIDC providers now show only `OIDCConfigurationInline`; SAML providers show RBAC/SAML inlines. Previously all inlines were appended unconditionally.
- `get_fields()` / `get_form()`: OIDC providers use a dedicated `OIDCSocialAppForm` that exposes `server_url` (read/written to `SocialApp.settings["server_url"]`) and `secret` (masked with `PasswordInput`). The `sites` M2M field and CSV upload fields are hidden.
- `save_model()`: auto-assigns `Site.objects.get_current()` for OIDC providers; skips CSV import logic.

**`identity_providers/forms.py`** — adds `OIDCSocialAppForm` with a `server_url` URLField and a masked `secret` widget.

**`identity_providers/migrations/0005`** — renames a `related_name` on `IdentityProviderUserLog` to avoid a reverse accessor clash introduced by the new `oidc_auth` models.

**`cms/settings.py`**:
- Adds `allauth.socialaccount.providers.openid_connect` and `oidc_auth` to `INSTALLED_APPS`.
- Sets `USE_OIDC = False` as the default flag.
- Sets `SOCIALACCOUNT_ADAPTER` to `UnifiedSocialAccountAdapter` as the new default.

**`files/context_processors.py`** / **`templates/config/core/user.html`** — exposes `USE_OIDC` to templates to conditionally show the SSO profile link.

---

### Tests

- `oidc_auth/tests/test_models.py` — model validation, uniqueness constraints
- `oidc_auth/tests/test_adapter.py` — adapter, `handle_ietf_role_mapping`, `sync_profile_picture`
- `identity_providers/tests/test_adapter.py` — `UnifiedSocialAccountAdapter` dispatch for SAML, OIDC, and unknown providers

Run with:
```bash
docker compose exec web pytest oidc_auth/tests/ identity_providers/tests/ -v
```

---

### Security notes

- `save_oidc_response_logs` (which persists raw OIDC claims) defaults to `False` and is accompanied by a warning in the model's `help_text` that it contains sensitive data.
- Username sanitisation in `populate_user` uses an allowlist regex (`[^\w.@-]`) consistent with MediaCMS's existing username validators, preventing injection of special characters from untrusted `sub` claims.

---

### Documentation

- `docs/oidc_setup.md` — Feature overview

---

### Files changed

| Path | Change |
|------|--------|
| `cms/settings.py` | INSTALLED_APPS, USE_OIDC, SOCIALACCOUNT_ADAPTER |
| `cms/dev_settings.py` | Enable USE_OIDC for local dev |
| `deploy/docker/nginx_http_only.conf` | OIDC callback location |
| `docker-compose.yaml` | Pass OIDC_CLIENT_SECRET to web service |
| `files/context_processors.py` | Expose USE_OIDC |
| `templates/config/core/user.html` | Conditional SSO profile link |
| `frontend/src/static/js/components/list-item/ListItem.jsx` | Show name |
| `identity_providers/adapter.py` | UnifiedSocialAccountAdapter (+103 lines) |
| `identity_providers/admin.py` | OIDC-aware inlines, fields, form, save_model |
| `identity_providers/forms.py` | OIDCSocialAppForm (server_url + masked secret) |
| `identity_providers/migrations/0005_*` | Rename related_name |
| `identity_providers/models.py` | related_name adjustment |
| `identity_providers/tests/test_adapter.py` | New (+320 lines) |
| `oidc_auth/` | New Feature (entire directory) |
| `docs/oidc_setup.md` | New |
| `docs/admins_docs.md` | Add the `oidc_setup.md` reference |

## Steps
*Pre-deploy*

Rebuild the frontend assets
To enable OIDC support, add configuration to local_settings.py

*Post-deploy*
Synchronize the OIDC configuration:
```
 python manage.py sync_oidc_providers
 ```


mgogoulos:
Thanks for opening the PR!

Couple of questions and notes from initial code review:

1. Are you already using this? If so in what identity provider, and also in what environment (eg university/organization)
2. Did you have the chance to run it along with SAML? Because I see the unified adapter...
3. I see the settings in local_settings.py, but also the configuration in Django admin. Did you test both ways or one of them? Also can you explain how this will work in the admin side? What changes are happening in the existing experience

4. There is some irrelevant stuff that will have to be justified (eg frontend/src/static/js/components/list-item/ListItem.jsx or deploy/docker/nginx_http_only.conf) . These changes make PR review harder, and we can't just let them without them being justified, the rist to solve something but introduce issues on other environments is big that way. Ideally, whatever change is not necessary here should be removed, to make code review easier too

5. Part of the AI code formatting will have to be removed/edited, eg lines with --------------------------------, too much commenting on parts that are trivial to read, imports in non top level places etc.
6. This is a big change overall that I'd like to try out myself, once the above are fixed, so I might need suggestion on the environment and some help...

Regards

GaijinKa:
Thanks! I'll address everything point by point.

> 1. Are you already using this? If so in what identity provider, and also in what environment (eg university/organization)

Yes, this is currently running with the IETF-provided IdP.

> 2. Did you have the chance to run it along with SAML? Because I see the unified adapter...

I didn't test SAML together with OIDC in this setup.

> 3. I see the settings in local_settings.py, but also the configuration in Django admin. Did you test both ways or one of them? Also can you explain how this will work in the admin side? What changes are happening in the existing experience

The configuration in `local_settings.py` is only used as a bootstrap reference. It is not meant to be the source of truth at runtime.

It is imported into Django admin via the management command `sync_oidc_providers`, which synchronizes the configuration into the database. Once imported, the full configuration can be managed and updated directly from the Django admin interface.

Regarding authentication behavior, existing users created with password-based authentication are not affected. For example, admin access via the standard user defined through environment variables still works as expected.

No other substantial changes are introduced to the existing authentication experience.
An example is documented in `docs/oidc_setup.md` under *Configure MediaCMS Settings*.

> 4. There is some irrelevant stuff that will have to be justified (eg frontend/src/static/js/components/list-item/ListItem.jsx or deploy/docker/nginx_http_only.conf) . These changes make PR review harder, and we can't just let them without them being justified, the rist to solve something but introduce issues on other environments is big that way. Ideally, whatever change is not necessary here should be removed, to make code review easier too

- The nginx change is needed to correctly build the redirect/callback URL when running behind non-standard ports. Without it, the IdP callback breaks in those cases.
- The change in `ListItem.jsx` is required because the OIDC `sub` claim is used as the username, and it can be numeric. This breaks UI assumptions that usernames are always string-like. The update ensures stable rendering and identity display.

And both changes are backward compatible. Feel free to point me any other change that you want me to explain to facilitate your review.

> 5. Part of the AI code formatting will have to be removed/edited, eg lines with --------------------------------, too much commenting on parts that are trivial to read, imports in non top level places etc.
I'll clean up formatting artifacts (separator lines, excessive comments, and other noise from autogenerated changes).

If it looks better for you, I can scratch the autogenerated tests at all

> 6. This is a big change overall that I'd like to try out myself, once the above are fixed, so I might need suggestion on the environment and some help...

I'm going to leave this answer to @alexamirante. He will provide more context on this PR and the one he submitted (#1530)

Regards!

alexamirante:
Thanks @mgogoulos! I've just created #1541 to hopefully shed some light on what we're trying to achieve.

I can help setting you up with the [test instance of the IETF Datatracker](https://dt-main.dev.ietf.org/), which is the IETF IdP.

PR #1546, opened 2026-06-25
feat: add opt-in TwelveLabs video analysis (transcript, description, tags)

mohit-twelvelabs:
Hi! I'm Mohit, I work at TwelveLabs (@mohit-twelvelabs).

## Description

This PR adds an **opt-in** integration with [TwelveLabs](https://twelvelabs.io) Pegasus video-understanding model, as an alternative to the existing Whisper transcription. In a single pass over an uploaded video it produces:

- a **transcript** (saved as a `.vtt` subtitle, like Whisper),
- a short **description** (filled in only when the uploader left the description empty), and
- a set of **tags** attached to the media.

Why it helps MediaCMS: it gives admins a one-call way to auto-enrich uploads with searchable metadata (transcript + description + tags) instead of just subtitles, improving accessibility and on-site search/discovery.

It mirrors the existing Whisper flow as closely as possible:
- a new per-media `allow_twelvelabs_analyze` flag enqueues a `twelvelabs_analyze` Celery task (same `long_tasks` queue, same `TranscriptionRequest` bookkeeping);
- **disabled by default** (`USE_TWELVELABS_ANALYZE = False`) so a default install behaves exactly as before — nothing runs and the dependency is not required unless you turn it on;
- the API key is read from the `TWELVELABS_API_KEY` environment variable, so it never lives in source;
- the `twelvelabs` package is added to `requirements-full.txt` (same place as `openai-whisper`), so the `mediacms:full` image / `docker-compose.full.yaml` already ships it;
- docs added under admin docs section 29.

This requires no maintainer-side credentials to review; it's purely additive and gated behind a setting.

### How it was tested
- Added a no-network unit test (`tests/test_twelvelabs.py`) for the response parser (markdown-fence stripping, tag trimming, invalid-JSON defaults) plus a live wiring test that is skipped unless `TWELVELABS_API_KEY` is set.
- Verified the TwelveLabs Python SDK contract (`assets.create` / `analyze` with `VideoContext_AssetId`) against `twelvelabs==1.2.8`, and confirmed live API connectivity with a Marengo text-embedding call returning a 512-dim vector. Full Pegasus video runs are wired and exercised via the SDK; long server-side processing time was not run end-to-end in CI.
- `flake8` clean on all changed files (repo config, max-line 119).

You can grab a free API key at https://twelvelabs.io — there's a generous free tier.

## Steps

*Pre-deploy*
- None for existing installs (feature is off by default).
- To enable: install `requirements-full.txt` (or use the `:full` image), set `USE_TWELVELABS_ANALYZE = True` and `TWELVELABS_API_KEY` in `local_settings.py`/env, and run `python manage.py migrate` (one additive `BooleanField`).

*Post-deploy*
- None.


mgogoulos:
thanks @mohit-twelvelabs for opening the PR, it looks valid and appreciate making it opt-in by default.

We'd better avoid adding a new field in Media model, and it's a good time to make it more generic, so that other services could be enabled and disabled in the future. Maybe introduce a new model for these type of services. Expect some feedback soon, feel free to write your thoughts as well

mohit-twelvelabs:
Thanks @mgogoulos — that direction makes a lot of sense, and I agree adding a one-off field to `Media` doesn't scale. Making this a generic, pluggable service layer is the right call.

A sketch of what I had in mind, to align before I refactor:

- A small `MediaService` (or `ServiceIntegration`) model describing an available service — e.g. `key` (`twelvelabs`), `name`, `enabled` (admin toggle), and optional JSON `config` — so services can be turned on/off without code changes.
- A generic per-media results model, e.g. `MediaServiceResult` with `media` (FK), `service` (FK/key), `status`, `result` (JSONField), `created_at` — instead of TwelveLabs-specific columns on `Media`. TwelveLabs transcript/description/tags just become one `result` payload; future services (other vision/transcription/moderation providers) slot in the same way.
- TwelveLabs becomes the first concrete provider behind a thin interface (`run(media) -> result dict`), kept opt-in/off by default as it is now.

That keeps `Media` untouched and gives you a clean extension point. Happy to align on naming/shape to match your conventions (e.g. whether you'd prefer a registry in settings vs a DB-backed `MediaService` model) — and I'll rework the PR along these lines once you confirm the approach. Looking forward to the feedback you mentioned!

mgogoulos:
Good suggestions. How about we create a new app `Integration` and place everything into `integrations`, even the new models (which I suggest it to be `Integration` instead of `MediaService`). So that would be `integrations/twelvelabs` for TwelveLabs.

There could be a list of integrations (TwelveLabs only for now) that can be extended when adding others, and thus when setting up TwelveLabs you need to add a new Integration from the admin and the rest options, which are a good baseline for any service:

- service (select TwelveLabs from a drop list)
- enabled
- model_name
- api_key
- json confir for rest

The actual code (eg `twelve_labs_utils.py`) can live in `integrations/twelvelabs`. No need to introduce new files for the celery tasks, they can be called defensively (without importing at the top level) as you do now. 

Regarding the python client, thanks for placing it under the full image. This still forces to install dependencies that might be unused in many cases, so a more clear solution has to be introduced... 
Besides the fact that the image will be getting bigger, there's the supply chain management security issues that are becoming very threatening these days. 

Idea, how about introducing a file requirements-optional.txt with the following content

twelvelabs==1.2.8
  # future optional services get pinned here too

that is not installed by Dockerfile, but instead there's something like a check in entrypoint.sh

  if [ "$ENABLE_TWELVELABS" = "True" ]; then
          echo "entrypoint.sh: installing TwelveLabs client..."
          uv pip install -c requirements-optional.txt twelvelabs
      fi


This will still require the `ENABLE_TWELVELABS = True` to be set in settings.py, and will introduce a latency when docker compose up is made but I think it's an acceptable way of getting optional services up and running, and also keep the ability to update version is a requirements.txt file instead of editing entrypoint.sh...

What we gain is the fact that extra dependencies are installed only when explicitly enabled. When the core configuration moves to Kubernetes this might require adjustments but this is something to deal in the future. 

I could also refactor Whisper to get to be installed like that, and move it's code from files/ and also remove the need of a seperate image (the full one). 

Hope these make sense and are somehow alligned with what you're working on. Your PR is in a very good state so thanks for getting to make these changes in order for it to become a good base for adding more services in the future!

mohit-twelvelabs:
Thanks @mgogoulos, this design makes a lot of sense — I've restructured the PR to follow it (pushed as 07e6b7f).

**New `integrations` app** (registered in `INSTALLED_APPS` as `integrations.apps.IntegrationsConfig`, mirroring the `lti` app layout) with a generic `Integration` model:

- `service` — choices field (`twelvelabs` the only option for now; new services just get added to the `Service` choices)
- `enabled` — boolean
- `model_name` — char (optional; falls back to a sensible default)
- `api_key` — char (rendered with a masked/password widget in the admin)
- `config` — `JSONField` for the rest (currently supports `prompt` and `max_tokens` overrides)
- registered in the admin, so setup is just "add an Integration" — no settings/env needed

**TwelveLabs code** moved to `integrations/twelvelabs/twelve_labs_utils.py`. It reads `api_key` / `model_name` / options from the enabled `Integration` row instead of a Media field or `settings.py`.

**Removed from before:** the `allow_twelvelabs_analyze` field + its migration on the `Media` model, and all the `TWELVELABS_*` / `USE_TWELVELABS_ANALYZE` settings. The `Media` model no longer references TwelveLabs at all.

**Celery:** no new task modules — the `twelvelabs_analyze` task stays in `files/tasks.py`. It's triggered from the existing upload flow (`Media.media_init`), and both the trigger and the task resolve the integration defensively and import the util lazily, so with no enabled `Integration` it's a complete no-op and the default install behaves exactly as before.

I also gave `service` a `unique=True` constraint (one row per service) so `get_active()` is unambiguous — happy to drop that if you'd prefer allowing multiple rows per service.

One thing I'd love your read on: do the field names/shape (`service` / `enabled` / `model_name` / `api_key` / `config`) match what you had in mind, and is `Integration` the right model name (vs. e.g. per-service subtyping later)? Easy to adjust — this is your call on the direction.

